PR #6, opened 2017-10-07
Added Angular Universal.

ahaw:
Start it with: npm run build:dynamic && npm run serve:dynamic.

### DIFF
--- a/.angular-cli.json
+++ b/.angular-cli.json
@@ -6,7 +6,7 @@
   "apps": [
     {
       "root": "src",
-      "outDir": "dist",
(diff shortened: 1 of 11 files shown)


tomastrajan:
Thank you for effort will have to review it later, don't have computer access at the moment.

ahaw:
On slower machine I noticed strange flickering during initial load.

tomastrajan:
Universal needs more complicated setup and does't just run out of the box which is a goal of this repo as a "starter" project so this will not happen. Sorry and still thank you for taking your time.

PR #7, opened 2017-10-08
Update according to last changes in Angular Material

Arelav:
Almost all Modules in Angular Material was renamed from MdModuleName to MatModuleName same as component selectors from md-component to mat-component


tomastrajan:
This was done on master and released already on 7.10. 2017 as version 1.15.0.

PR #8, opened 2017-10-08
chore(codestyle): fix theme filename and reference typo

Arelav:
Typo fix - sccs seems to be scss according to other components.


tomastrajan:
Also branch name is then chore/fix-typo.

tomastrajan:
Also adjusted build so that it doesn't try to deploy PRs so that your build is green now !

PR #9, opened 2017-10-14
chore(codestyle): fix theme filename and reference typo

Arelav:


tomastrajan:
Hey man! Thank you for your effort but sorry it didn't work. The commit message is still Fix typo if you check in commit tab and also it is impossible to merge now because it fell behind in history. 

It is a small change so it maybe makes sense to do once again from start correctly. If you don't want I can just do it myself but I would prefer if it was you because it would be cool to have a contributor !

Arelav:
OK. I'll try again. I'm not very familiar with GitHub flow. This is one of reasons of this pull request, I want to try how it works :)

tomastrajan:
Congratulations! It worked 😉 I am also in process of learning this whole github open source thing, both as commiter and maintainer.  Thank you again for your congribution and have a great day!

PR #36, opened 2018-01-13
fix(todos): eliminate input mutation from the reducer

alexkonovalov:
*Problem:* the changed items in the array are not being recreated. Instead, [the shallow copy of ```items``` object is being produced](https://redux.js.org/docs/recipes/reducers/ImmutableUpdatePatterns.html#updating-nested-objects) and they are being mutated by  ```item.done = !item.done ```. 

*Solution:* use ```map``` to produce a new array and create a new object for every toggled/untoggled item.

### DIFF
--- a/src/app/examples/todos/todos.reducer.ts
+++ b/src/app/examples/todos/todos.reducer.ts
@@ -71,14 +71,11 @@ export function todosReducer(state = initialState, action: TodosActions) {
       });
 
     case TodosActionTypes.TOGGLE:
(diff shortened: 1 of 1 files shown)


tomastrajan:
Great catch, thank you! Coukd you please change commit to fix instead of feat🙏? Then I'll hapilly merge 😉

alexkonovalov:
@tomastrajan done!

tomastrajan:
@alexkonovalov thank you, merged! 🎉

PR #48, opened 2018-02-01
Fix E2E Tests

peterkrieg:
Currently, when you clone from master and run `ng e2e` you get the error:  `Failed: No element found using locator: By(css selector, app-root h1)`. 

This simple PR just fixes the e2e test to find the correct element and expect a different value of text.

### DIFF
--- a/e2e/app.e2e-spec.ts
+++ b/e2e/app.e2e-spec.ts
@@ -9,6 +9,6 @@ describe('angular-ngrx-material-starter App', () => {
 
   it('should display message saying app works', () => {
     page.navigateTo();
-    expect(page.getParagraphText()).toEqual('app works!');
+    expect(page.getParagraphText()).toEqual('ANGULAR NGRX MATERIAL STARTER');
   });
 });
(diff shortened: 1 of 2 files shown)


tomastrajan:
Thank you very much for your effort. Could you please change commit message to `fix(e2e): correct message and selector` or something like that? We're generating changelog automatically so the commits should follow [standardized format](https://github.com/tomastrajan/angular-ngrx-material-starter/blob/master/CONTRIBUTING.md).

peterkrieg:
Updated the commit message, sorry for not taking a look at contributing guide beforehand.  Awesome boilerplate by the way!  Will be considering using parts of this for my team's next project.

tomastrajan:
No worries! I'll merge PR once the CI build is completed. Happy to hear that you find this useful. I am always trying to improve it a bit here and there. Should definitely add more tests, both unit and e2e though...

tomastrajan:
In a near future I want to add custom schematics for generating themed components with `some.component.scss-theme.scss` stub files generated and automatically added to the main styles.scss.

peterkrieg:
Yeah it's super useful, and I also really enjoyed reading your medium post about it.  Thanks for making it!

PR #138, opened 2018-04-12
fix(examples): fix stock market API source

shootermv:
google stock api replced by "iextrading" one

### DIFF
--- a/src/app/examples/stock-market/stock-market.service.ts
+++ b/src/app/examples/stock-market/stock-market.service.ts
@@ -21,8 +21,10 @@ export class StockMarketService {
           last: stock.latestEPS,
           ccy: 'USD',
           change: stock.c,
-          changePositive: stock.day30ChangePercent.toString().indexOf('+') === 0,
-          changeNegative: stock.day30ChangePercent.toString().indexOf('-') === 0,
+          changePositive:
+            stock.day30ChangePercent.toString().indexOf('+') === 0,
+          changeNegative:
+            stock.day30ChangePercent.toString().indexOf('-') === 0,
           changePercent: stock.day30ChangePercent.toFixed(2)
         }))
       );


src/app/examples/stock-market/stock-market.service.ts
```diff
@@ -6,30 +6,26 @@ import { map } from 'rxjs/operators';
...
+      .get(PROXY_URL + `https://api.iextrading.com/1.0/stock/${symbol}/stats`)
       .pipe(
-        map((res: string) => JSON.parse(res.replace('//', ''))[0]),
         map((stock: any) => ({
-          symbol: stock.t,
-          exchange: stock.e,
-          last: stock.l,
+          symbol: stock.companyName,
+          exchange: stock.EBITDA,
+          last: stock.latestEPS,
```
tomastrajan: Unfortunately this is not correct. I tried it with `FB` which is currently around 163 and the `latestEPS` property value is `5.93` also calculating change should be done from the last price not a 30 day average.

As I looked into API response, it unfortunately doesn't provide the numbers we need.

The API can be delayed or not precise, the exact numbers are not important nut it would be very nice if they were at least roughly realistic.

tomastrajan:
Hi @shootermv !

Thank you very much for your effort! Unfortunately the data source doesn't seem to provide the data we need ( see the code review comment ). So at the time I can't merge this PR. Anyway I hope we can make this work ;)

Cheers!

ps: it is also good practice and I do it when I create PR to other repos to squash all commits into single commit which is then used to create pull request. I usually use `git rebase -i HEAD~<number of commits to squash>`.

tomastrajan:
ps2: it is also great to add `, fixes #<issue number>` to the end of commit message to automatically link it to the Github issue ! ;)

tomastrajan:
This was already fixed by other PR.

Issue #178, opened 2018-04-29
Improve / add more e2e test


tomastrajan:
This repository currently runs single e2e test as a part of it's CI (continuous integration) build on Travis CI. Test can also be executed using `npm run e2e`.

It would be great to provide more e2e test examples for the users of this repository to serve as a great example for how to write their own tests.

- best practices for angular e2e testing ( eg: separate tests and page objects ... )
- more e2e test (mostly for the examples functionality) like add todo, toggle todo, ...
- suggest / evaluate / use community driven e2e test library for angular

Tests can be added in step by step fashion with one PR corresponding to the tests of one route / page to keep the required effort low but please feel free to add more ;)

ps: please use commit message in format like feat(e2e): add <test-description > tests, #<issue-number>, just replace test description with stuff like "todos" based on applicable route / page.

christophechevalier:
In my app, I use this front-end testing tool : 

> https://www.cypress.io/

tomastrajan:
Closing this issue because it is too generic, will open additional issues (one per component / feature) to better track progress and to introduce ability to close issues with the commit message.

PR #179, opened 2018-04-29
feat(e2e): #178 add tests  for about page

shootermv:
Hi,  this is my first try, please correct me if something wrong

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="6.0.0-rc.4"></a>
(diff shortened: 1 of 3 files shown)


tomastrajan:
Hey @shootermv !

First of all thank you for your effort. As you might be aware, there is already existing "app" e2e test which tests for the main heading of the about page similarly to the test you are adding.

My suggestion would be to test the "redirect" behaviour in the app e2e test and the heading in the about e2e test (so remove it from the app test). App test then can use additional checks for stuff like year in the footer. About test can contain stuff like presence of the main content like getting started and call to action buttons.

So tu summarize:

app e2e test
 - redirect behaviour
 - remove about h1 check
 - check for year in footer ?
 - check for menu ?

about e2e test
  - check for h1 content
  - check for main content like getting started and call to action buttons
  - remove redirect check

e2e utils (optional)
  - create e2e utils for stuff which might be needed on multiple occasions like checking for route url

Hope that this make sense and I am open to any suggestions and improvements.

Cheers!

PR #181, opened 2018-05-01
feat(e2e): #178 add tests for app and about page

shootermv:


tomastrajan:
Thank you for the PR! The tests are great! 

The only thing left is to use consistent simple arrow functions when only single expression is used. 
Example of simple arrow function is to use `.map(param => param.method())` instead of `.map(param =? { return param.method() })`.

Hope you don't mind and thank you again!

tomastrajan:
This was already merged from another PR.

PR #184, opened 2018-05-04
change e2e tests to use "async await" syntax

shootermv:
inspired by Vi Qi`s video you recommended, i think it will be better to use async await... what are you think?
Also, hope the branch/commit name are fit the convention (if not tell me i will modify)

### DIFF
--- a/e2e/src/about/about.e2e-spec.ts
+++ b/e2e/src/about/about.e2e-spec.ts
@@ -6,24 +6,21 @@ describe('About Page', () => {
 
   beforeEach(() => (page = new AboutPage()));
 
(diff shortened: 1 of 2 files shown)


tomastrajan:
Thank you for your effort! Will have to look bit more into this async await with protractor e2e tests.

In the meantime you can have a look into [angular conventional commit message](https://gist.github.com/stephenparish/9941e89d80e2bc58a153) to learn more about how to write commit messages that can be turned into changelogs automativally.

In case of this PR it could be somethibg like `refactor(e2e): use async await syntax`.

Will check and comment on the actual imolementation later.

Issue #177, opened 2018-04-29
Improve / add more unit tests


tomastrajan:
Currently repository contains only test stubs generated by Angular CLI (for the corresponding components and services).

These test stubs contains only necessary imports so that the basic `expect(component).toBeTruthy();` style test passes.

We need to improve tests to serve as a example for the users of this repository. 

- suggest best test practices
- standardized imports for test related stuff
- add unit tests for the component / service functionality
- use community driven test library? (optional)

Tests can be added in step by step fashion with one PR corresponding to the tests of one component / service to keep the required effort low but please feel free to add more ;)

ps: please use commit message in format like `feat(examples): add tests,  #<issue-number>`, just replace `examples` with applicable application part.

tomastrajan:
Closing this issue because it is too generic, will open additional issues (one per component / feature) to better track progress and to introduce ability to close issues with the commit message.

PR #242, opened 2018-06-12
feat(todo): create tests for todos container [added more TODOS] (#219)

shootermv:
added all "TODO"s expect the first one - cannot get menu show the menupopup

### DIFF
--- a/src/app/examples/todos/todos.component.spec.ts
+++ b/src/app/examples/todos/todos.component.spec.ts
@@ -12,6 +12,7 @@ import { SharedModule } from '@app/shared';
 import { TestStore } from '@testing/utils';
 
 import { TodosComponent } from './todos.component';
+import { ActionTodosToggle } from './todos.reducer';
 import {
   ActionTodosAdd,
   ActionTodosRemoveDone,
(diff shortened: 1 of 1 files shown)


tomastrajan:
Hi @shootermv !

Thank you for your PR, I added missing test and made minor adjustments to your test in [this commit](https://github.com/tomastrajan/angular-ngrx-material-starter/commit/e2535548a5f86512bfb200bd538d52a3d51d2399) , feel free to check it out how the filtering could have been solved.

Besides that I will close the original add tests issue. The issue is too generic which is not good for tracking of progress. 

I will create multiple new issues (one per feature / component) with the tests strings (similar to ones in the todos file) instead. That we can close these issues with pull requests that belong just to them and track progress.

Besides that, please use commit message like `test(todos): add tests` instead of `feat`. Its from the conventional commit standard [described here](https://github.com/conventional-changelog-archived-repos/conventional-changelog-angular/blob/master/convention.md). 

Wish you a great day!

tomastrajan:
Actually because in the future we will close issues with the commits the commit message should be more like this `test(<feature-name>): add tests, closes #<issue-number>` so for example `test(todos): add tests, closes #177`.

Cheers!

PR #280, opened 2018-07-08
feat(dependencies): migrate to the latest version of Angular, Ngrx, M…

hhubik:
…aterial, CLI

Angular 6.0.7
Ngrx 6.0.1
Material 6.3.2
CLI 6.0.8
RxJS 6.2.1
and a few other dependencies.


hhubik:
I want to add one of the tests listed in Issues, but I thought I would do this first.

tomastrajan:
@hhubik yeah its ok! You would get same versions even without bumping them when running local `npm install` because of the way semver treats `^` and `~` in version strings but your change will have impact on the versions in `package-lock.json` and hence on what gets installed using `npm ci` which is used in our continuous integration build on Travis CI.

PR #284, opened 2018-07-19
test(settings): Added settingsReducer unit tests

### LINKED ISSUE #269: Create unit tests for settings reducer
Create tests for `settings.reducer.ts` in `settings.reducer.spec.ts` file either by following [official ngrx testing guide](https://github.com/ngrx/platform/blob/master/docs/store/testing.md) or feel free to suggest any improvements.

hhubik:
closes #269

### DIFF
--- a/src/app/settings/settings.reducer.spec.ts
+++ b/src/app/settings/settings.reducer.spec.ts
@@ -1,0 +1,56 @@
+import {
+  initialState,
+  settingsReducer,
+  ActionSettingsChangeTheme,
+  ActionSettingsChangeAnimationsPage,
+  ActionSettingsChangeAnimationsPageDisabled,
+  ActionSettingsChangeAnimationsElements,
+  ActionSettingsChangeAutoNightMode
+} from './settings.reducer';
+
+describe('SettingsReducer', () => {
+  it('should return default state', () => {
+    const action = {} as any;
+    const state = settingsReducer(undefined, action);
+    expect(state).toBe(initialState);
+  });
+
+  it('should update theme', () => {
+    const action = new ActionSettingsChangeTheme({ theme: 'dark' });
+    const state = settingsReducer(undefined, action);
+    expect(state.theme).toEqual('dark');
+  });
+
+  it('should update pageAnimations', () => {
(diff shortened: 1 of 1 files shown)


src/app/settings/settings.reducer.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    const state = settingsReducer(undefined, action);
+    expect(state.pageAnimations).toEqual(false);
+  });
+
+  it('ActionSettingsChangeAnimationsPageDisabled should return state with updated pageAnimationsDisabled', () => {
+    const action = new ActionSettingsChangeAnimationsPageDisabled({
+      pageAnimationsDisabled: true
+    });
+    const state = settingsReducer(undefined, action);
+    expect(state.pageAnimationsDisabled).toEqual(true);
```
tomastrajan: If you check implementation of the reducer this action actually does two things:

1. sets `pageAnimationsDisable` to `true`
2. sets `pageAnimations` to `false`

Both are important for the functionality. 

Background: page animations are disabled in IE and EDGE because they don't work, disabling is not enough though, they have to be turned off too. Disabling is there just to prevent user from re-enabling them in settings.

src/app/settings/settings.reducer.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  ActionSettingsChangeTheme,
+  ActionSettingsChangeAnimationsPage,
+  ActionSettingsChangeAnimationsPageDisabled,
+  ActionSettingsChangeAnimationsElements,
+  ActionSettingsChangeAutoNightMode
+} from './settings.reducer';
+
+describe('settingsReducer', () => {
+  it('ActionSettingsPersist with undefined new state should return the initial state', () => {
+    const action: ActionSettingsPersist = {
```
tomastrajan: This action doesn't have reducer handler so its basically as testing `reducer(undefined, anythingRandom)` which would also trigger `default` reducer handler. As so I guess this test can be replaced with

```typescript
it('should return the default state', () => {
    const action = {} as any;
    const state = todosReducer(undefined, action);

    expect(state).toBe(initialState);
  });
```

because it's not specific to that persist action.

PR #293, opened 2018-08-09
test(settings): added SettingsEffects unit test

### LINKED ISSUE #270: Create unit tests for settings effects
Create tests for `settings.effects.ts` in `settings.effects.spec.ts` file either by following [official ngrx testing guide](https://github.com/ngrx/platform/blob/master/docs/store/testing.md) or feel free to suggest any improvements.

Try to keep tests as independent as possible (eg minimize importing of various modules in setting up tests)

hhubik:
closes #270


hhubik:
Hi @tomastrajan let me know what you think.

tomastrajan:
@hhubik btw I added @timdeschryver  as a maintainer of this repo, he is a ngrx team member and will help us to follow latest best practices for using NgRx!

src/app/settings/settings.effects.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        }
+      ]
+    });
+
+    localStorageService = TestBed.get(LocalStorageService);
+    animationsService = TestBed.get(AnimationsService);
+    effects = TestBed.get(SettingsEffects);
+  });
+
+  it('should not dispatch any action', () => {
```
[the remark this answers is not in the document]
timdeschryver: The reason why I would put it inside the block is because it's still testing the decorator of `persistSettings`.
If another effect would be added it would seem weird to me that this test is outside of the `persistSettings` block.

But after reading your explanation I'm also getting your point 😄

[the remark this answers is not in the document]
hhubik: @timdeschryver  OK. All tests should be in one block now.

src/app/settings/settings.effects.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    metadata = getEffectsMetadata(effects);
+    expect(metadata.persistSettings).toEqual({ dispatch: false });
+  });
+
+  it('should call methods on AnimationsService and LocalStorageService for PERSIST action', () => {
+    const action = {
```
timdeschryver: A nitpick, do we want to create these action "manually" or would it be better to use the action creators? @tomastrajan

tomastrajan: @timdeschryver @hhubik I would go with the action creators so `new SomeAction()`, actions and effects are tightly coupled anyway so they can be coupled in tests too, ideas?

timdeschryver: Yea I agree.

hhubik: @timdeschryver changed to use action creators.

PR #296, opened 2018-08-12
Update NgRx

timdeschryver:
This PR updates NgRx to version 6.1, therefor I also updated the angular deps because there were some unmet peer dependencies.

The `store.select` and `store.ofType` operators are deprecated now and will be removed in the next major release. Thus I already refactored those to use the pipeable operator. Because of this I also had to modify the existing `TestStore`, as reference I took the proposed version from ngrx/platform/pull/1027
.


tomastrajan:
@timdeschryver amazing work thank you very much for this contribution!

PR #298, opened 2018-08-14
refactor(examples): refactor ngrx selectors and todos state

tomastrajan:
This is a effort to bring ngrx usage more to official best practices, mainly with `createSelector` and `createFeatureSelector` functionality.

The main things are new `examples.state.ts` and `todos.selectors.ts` other stuff was more or less just shuffled around...


tomastrajan:
@timdeschryver yeah I am also not too happy with the `settings/settings` or `todos/todos`, let's think about it because also in case that we use `components` or `container` we would still need to introduce additional structure once there is more than one component. What about `todos/ui/todos-container` `todos/ui/todo-item` or maybe it would be better to use `todos/components/todos-container` ?

src/app/examples/examples.state.ts
```diff
@@ -0,0 +1,24 @@
+import { createFeatureSelector, ActionReducerMap } from '@ngrx/store';
+
+import { todosReducer } from './todos/todos.reducer';
+import { TodosState } from './todos/todos.model';
+import {
+  stockMarketReducer,
+  StockMarketState
+} from './stock-market/stock-market.reducer';
```
tomastrajan: This was the case why I had to split stuff into separate files. This file is necessary and has to import all reducers and all `<something>State` interfaces to be able to define whole feature module state interface and feature selector.

On the other hand we need feature selector in the reducers to be able to create specific selectors. That caused circular dependency so that the selectors and interfaces had to be extracted into separate files. I don't really see the way out of this, ideas @timdeschryver ?

src/app/examples/examples.state.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const FEATURE_NAME = 'examples';
+
+export const reducers: ActionReducerMap<ExamplesState> = {
+  todos: todosReducer,
+  stockMarket: stockMarketReducer
+};
+
+export const selectExamples = createFeatureSelector<any, ExamplesState>(
```
tomastrajan: Also here, we're in lazy loaded feature module and we're defining its state interface  and feature selector which accepts `AppState` and `FeatureState`. I am not completely sure what would happen if I imported `AppState` interface across the lazy boundary, would it be that just the interface gets included but bundles stay the same without some crazy duplication? I guess yes but haven't tried it yet...

Also the other way around, how can we get hold of this `ExamplesState` interface in `app.model.ts` without pulling in the rest of lazy loaded module content.

Ideas / experience with this @timdeschryver ?

src/app/examples/examples.state.ts
timdeschryver: How the NgRx example app is structured (and how we do it at work) is to create the selectors inside `examples.state`. This because you could use selectors from different "reducers", in this project this won't be the case (you won't combine a todo selector with a stock selector). The downside to this approach is that you can get a bloated file which is hard to maintain, because the app/module itself itsn't split up correctly (seperations of concerns).

I might be wrong but if we do this we might put everything in one file again, but on the other hand it feels better to create different files in this project as it will help starters understand the structure better imho.

timdeschryver:
My vote goes to `todos/components/todos-container`, the other one feels to deep for me personally.

PR #305, opened 2018-08-18
refactor(Core): Add App state

### LINKED ISSUE #304: Create proper AppState
Create `AppState` and use it for:
* stores in the app as in `Store<AppState>`
* createState helper in tests
* somewhere else ?

Verify that imports of interfaces across lazy loaded module boundaries don't cause unexpected content in the bundles. Verify with `npm run analyze`.

timdeschryver:
This PR adds an `AppState` to the application.
In order to make this possible I also split up the `auth` reducer, as @tomastrajan previously did for the example states.

What still has to be done is the verification that there is no unexpected appstate in the feature module's bundle with `npm run analyze`. I have taken a look and I think it should be OK, but please do verify it because this was new to me 😅 .

Closes #304

### DIFF
--- a/src/app/core/auth/auth.effects.ts
+++ b/src/app/core/auth/auth.effects.ts
@@ -9,9 +9,10 @@ import { LocalStorageService } from '../local-storage/local-storage.service';
 import {
   ActionAuthLogin,
   ActionAuthLogout,
-  AUTH_KEY,
   AuthActionTypes
-} from './auth.reducer';
+} from './auth.actions';
+
+export const AUTH_KEY = 'AUTH';
 
 @Injectable()
 export class AuthEffects {
--- a/src/app/core/core.state.ts
+++ b/src/app/core/core.state.ts
@@ -4,30 +4,32 @@ import {
   createFeatureSelector
 } from '@ngrx/store';
 import { storeFreeze } from 'ngrx-store-freeze';
+
+import { environment } from '@env/environment';
+
 import { initStateFromLocalStorage } from './meta-reducers/init-state-from-local-storage.reducer';
 import { debug } from './meta-reducers/debug.reducer';
 import { AuthState } from './auth/auth.models';
 import { authReducer } from './auth/auth.reducer';
-import { environment } from '@env/environment';
 
-export const reducers: ActionReducerMap<State> = {
+export const reducers: ActionReducerMap<AppState> = {
   auth: authReducer
 };
 
-export const metaReducers = (): MetaReducer<State>[] => {
-  // tslint:disable-next-line:prefer-const
-  let metas = [initStateFromLocalStorage];
-  if (!environment.production) {
-    metas.unshift(storeFreeze);
-    if (!environment.test) {
-      metas.unshift(debug);
-    }
+export const metaReducers: MetaReducer<AppState>[] = [
+  initStateFromLocalStorage
+];
+if (!environment.production) {
+  metaReducers.unshift(storeFreeze);
+  if (!environment.test) {
+    metaReducers.unshift(debug);
   }
-  return metas;
-};
+}
 
-export const selectAuthState = createFeatureSelector<State, AuthState>('auth');
+export const selectAuthState = createFeatureSelector<AppState, AuthState>(
+  'auth'
+);
 
-export interface State {
+export interface AppState {
   auth: AuthState;
 }
--- a/src/app/core/core.state.ts
+++ b/src/app/core/core.state.ts
@@ -1,0 +1,33 @@
+import {
+  ActionReducerMap,
+  MetaReducer,
+  createFeatureSelector
+} from '@ngrx/store';
+import { storeFreeze } from 'ngrx-store-freeze';
+import { initStateFromLocalStorage } from './meta-reducers/init-state-from-local-storage.reducer';
+import { debug } from './meta-reducers/debug.reducer';
+import { AuthState } from './auth/auth.models';
+import { authReducer } from './auth/auth.reducer';
+import { environment } from '@env/environment';
+
+export const reducers: ActionReducerMap<State> = {
+  auth: authReducer
+};
+
+export const metaReducers = (): MetaReducer<State>[] => {
+  // tslint:disable-next-line:prefer-const
+  let metas = [initStateFromLocalStorage];
+  if (!environment.production) {
+    metas.unshift(storeFreeze);
+    if (!environment.test) {
+      metas.unshift(debug);
+    }
+  }
+  return metas;
+};
+
+export const selectAuthState = createFeatureSelector<State, AuthState>('auth');
+
+export interface State {
+  auth: AuthState;
+}
--- a/src/app/examples/examples.state.ts
+++ b/src/app/examples/examples.state.ts
@@ -1,5 +1,7 @@
 import { createFeatureSelector, ActionReducerMap } from '@ngrx/store';
 
+import { AppState } from '@app/core';
+
 import { todosReducer } from './todos/todos.reducer';
 import { TodosState } from './todos/todos.model';
 import { stockMarketReducer } from './stock-market/stock-market.reducer';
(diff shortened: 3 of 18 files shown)


tomastrajan:
@timdeschryver cool! Will have a look and let you know, thank you!

tomastrajan:
Hi @timdeschryver! The output of the `npm run analyze` look alright, the import definitely didn't pull whole `CoreModule` into lazy bundle (`ExamplesModule`) so that's great! I will add some more comments into the code shortly.

src/app/examples/examples.state.ts
```diff
@@ -12,11 +14,15 @@ export const reducers: ActionReducerMap<ExamplesState> = {
...
+export const selectExamples = createFeatureSelector<State, ExamplesState>(
   FEATURE_NAME
 );
 
 export interface ExamplesState {
   todos: TodosState;
   stockMarket: StockMarketState;
 }
+
+export interface State extends AppState {
```
tomastrajan: I would like to understand more about the tradeoff of this solution. This seems to lead to "re-creating" application state in every lazy loaded module. What would happen if we defined `examples` property in the original `AppState` ? Maybe it leads to circular deps...

Or is this the official ngrx way? If so maybe a bit change of naming ? Like `ExtendedAppState` or something that signifies what we're dealing with.

src/app/examples/stock-market/components/stock-market-container.component.spec.ts
```diff
@@ -172,5 +173,5 @@ function createState(stockState: StockMarketState) {
     examples: {
       stockMarket: stockState
     }
-  };
+  } as State;
```
tomastrajan: `as State` vs `:State` what is the tradeoff here ?

src/app/core/auth/auth.effects.ts
```diff
@@ -9,9 +9,9 @@ import { LocalStorageService } from '../local-storage/local-storage.service';
 import {
   ActionAuthLogin,
   ActionAuthLogout,
-  AUTH_KEY,
   AuthActionTypes
-} from './auth.reducer';
+} from './auth.actions';
+import { AUTH_KEY } from './auth.reducer';
```
tomastrajan: Let's movie `AUTH_KEY` to this file so that import is not necessary ( was done similarly for todos, stocks, ... )

src/app/core/core.state.ts
```diff
@@ -0,0 +1,35 @@
...
+  initStateFromLocalStorage
+];
+if (!environment.production) {
+  metaReducers.unshift(storeFreeze);
+  if (!environment.test) {
+    metaReducers.unshift(debug);
+  }
+}
+
+export const selectAuthState = createFeatureSelector<AppState, AuthState>(
```
tomastrajan: Currently we have only `AuthState` in here but in theory there can be more reducers and related states. This would be then more like `selectCoreState` which has auth state in it similarly like `selectExampleState` has `todos` and `stocks` in it.

tomastrajan: We can also thing about moving settings into core. The setting was implemented as a stand alone module with component, reducer, etc but is eagerly loaded so that's the question then. Does every eagerly loaded module have to be in core or can they be extracted away. Maybe it only has to be in core if any other core stuff needs its functionality. If not it can be extracted into stand alone eager module like it is now.

src/app/examples/examples.state.ts
```diff
@@ -1,5 +1,7 @@
 import { createFeatureSelector, ActionReducerMap } from '@ngrx/store';
 
+import { AppState } from '@app/core/core.state';
```
tomastrajan: In `CoreModule` we're still using `index.ts` file which re-exports everything so we can simplify this import as `'@app/core'`. Maybe we will move away from `index.ts` barrels in the future. They were promoted in the beginning, then not and now I think they are still promoted for `library` style packages with `public_api.ts` or something similar.

src/app/examples/examples.state.ts
timdeschryver: I followed the NgRx example-app. 
But in our case this "ExtendedState" has little to no benefit.
If you would ever need something from the `AppState` in your feature module's selectors, this would be the way to make your feature module "smart" enough to know what's inside the AppState's model. 

We can remove this, the only impact it has is that we won't be able to safely type our `createFeatureSelector`, this would look as the following:

```
export const selectExamples = createFeatureSelector<ExamplesState>(FEATURE_NAME);
```

Personally I don't really have an opinion on this, this may help beginners, or this could be confusing them.

src/app/examples/stock-market/components/stock-market-container.component.spec.ts
timdeschryver: If you would do `:State` you would have to create your whole State tree, so this would become:


```ts
return {
    examples: {
      stockMarket: stockState,
      todos: {} as TodosState
    },
    auth: {} as AuthState
  };
```

Because we're only interested in the `stockMarket` here, I didn't really like this way. Also if you would add another example state, your tests would fail while the change doesn't impact these tests.

We could create a helper to built the entire state, and we would use this helper to built our entire state in every test?

src/app/core/core.state.ts
timdeschryver: Exactly what you're saying.
> Maybe it only has to be in core if any other core stuff needs its functionality. If not it can be extracted into stand alone eager module like it is now.

src/app/examples/examples.state.ts
timdeschryver: I think it's also used with the settings and shared module.
I think with the `paths` defined in `tsconfig.json`, we don't need these `index` files anymore?

tomastrajan: In settings we have for example in `settings.effects.ts` import like this `import { LocalStorageService, AnimationsService } from '@app/core';` 

I think that for core it's also pretty nice, we're not interested in the internal structure of core so we're fee to shuffle things around inside as long as the exported services still exists. 

Other modules / services usually import only their own stuff which is always imported relatively and besides that they consume `CoreModule` so in that case `index.ts` and short import paths can make sense hmm?

tomastrajan: Also now I realized, I forgot to refactor settings to split out selectors / actions / etc... But will do it after this is finished :)

src/app/core/core.state.ts
tomastrajan: Yeah, let's go with that. Still, the previous comment talks about `selectCoreState ` vs `selectAuthState`, please have a look and let me know what do you think.

src/app/examples/stock-market/components/stock-market-container.component.spec.ts
tomastrajan: Thank you for the explanation! The solution then seems great because it enables us to ignore changes to other modules and their state which is irrelevant for the particular test. Let's not go with entire state helper so that the tests and their setup is as isolated as possible! 👍

src/app/examples/examples.state.ts
tomastrajan: I see, thank you for the explanation! Let's keep it! I can imagine use cases when we might need `auth` or `settings` stuff from store also in other feature modules. I will try to come up with such example and implement it so that it is used for real in our example app. 👍

src/app/examples/examples.state.ts
timdeschryver: After reading your comment it makes sense to have an `index` for `CoreModule` 👍

src/app/core/core.state.ts
timdeschryver: I think `selectAuthState` is better in this case, we are selecting `auth` from the `AppState`.
If for example settings would also be part of the `AppState` we would have another selector for settings `export const selectSettingState = createFeatureSelector<AppState, SettingsState>( 'settings');`

timdeschryver: does that make sense what I told here? Or should we go for `selectCoreState`?

tomastrajan: @timdeschryver yes sure! I just was a bit tight on time last couple of days cuz I am visiting family in Slovakia :) Will prolly merge now as I already did for `lint-staged` :)

timdeschryver: Sorry, wasn't meant to be a "Hey merge me" message 😅 
Enjoy your time in Slovakia!

tomastrajan: No worries, was not taken like that :) Thanks for the great PR and discussion about the concepts!

PR #306, opened 2018-08-19
chore: add lint-staged to only format modified files

timdeschryver:
When we make a commit all files are getting formatted with prettier, even the unmodified ones.
Also after they have been formatted they won't automatically be added to the commit.

This PR changes this behavior and will only format the modified files and will also add the formatted file to the commit , following the [prettier guide](https://prettier.io/docs/en/precommit.html#option-1-lint-staged-https-githubcom-okonet-lint-staged).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -830,6 +830,15 @@
         "tslib": "^1.9.0"
       }
     },
+    "@samverschueren/stream-to-observable": {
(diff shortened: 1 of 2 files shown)


tomastrajan:
Hi @timdeschryver! This is great, thanks!

PR #310, opened 2018-08-23
feat(NgRx): add store devtools

timdeschryver:


tomastrajan:
Would it also make sense to only add dev tools conditionally only in DEV environment?

timdeschryver:
Yea, we can do that too.
I'll change it later today.

tomastrajan:
@timdeschryver perfect! Thank you!

Issue #315, opened 2018-08-28
stock-market refresh produce error stock undefined.


cloudstr21:
Hi there,

I founded an issue when i open examples page => stock-market sub page.
How to reproduce the issue?
- open stock market page
- refresh the page
- an error will occurred, it is says "stock undefined" as shown picture below.
![issue ngrx](https://user-images.githubusercontent.com/13480230/44696183-aa182b00-aaa0-11e8-83e4-68c4f556b0f2.png)

Thanks

timdeschryver:
Seems like you're right, thanks for letting us know!
Do you want to open up a PR with a fix?

tomastrajan:
@cloudstr21 @timdeschryver  this only happens if navigating / refreshing directly on the `stock-market` url, probably is a race condition with initializing state from local storage / reducer initial state and expecting that state in the `StockMarketComponent` `ngOnInit` method.

cloudstr21:
Sure. I will make it as soon as I found the solutions.

timdeschryver:
Awesome, let us know if we can help in any way possible.

cloudstr21:
this is strange, select data from ngrx/store is always undefined after page is refreshing. @timdeschryver
i tried to set timeout, but the result is the same.
`this.store.pipe(select(selectStockMarket))`. selector, effects, etc is fine, and this error doesn't happens in todos component.

timdeschryver:
The problem lies with the `initStateFromLocalStorage` meta-reducers.
Our state shape looks like the following: 
````
{
  todos: {…}, 
  stockMarket: {…}
}
```

But when we init our app via `initStateFromLocalStorage`, we get the following state:
````
{
  todos: {…}, 
  stocks: {…} // this should be stockMarket
}
```

timdeschryver:
To fix this we have two choices:

- rename `stockMarket` to `stocks`
- rename modify `STOCK_MARKET_KEY` in `stock-maerket.effects` to be `EXAMPLES.STOCKMARKET` - but this still won't work because in `LocalStorageService.loadInitialState` it will transform this string to lowercase, making it `stockmarket` instead of `stockMarket`.

timdeschryver:
I forgot to mention, that I would vote to rename `stockMarket` to `stocks`

cloudstr21:
That's it. you are great @timdeschryver . I just fix it as your suggestions. now it run perfectly in my local. Should i re-commit it?

[...]

PR #316, opened 2018-08-28
Update stock-market-container.component.ts #315

cloudstr21:
I updated this file to occupied an error that occurred when refresh or navigate to stock-market link. I am not sure this is the best solutions or not, please take a review.

Thanks

### DIFF
--- a/src/app/examples/stock-market/components/stock-market-container.component.ts
+++ b/src/app/examples/stock-market/components/stock-market-container.component.ts
@@ -5,7 +5,10 @@ import { takeUntil } from 'rxjs/operators';
 
 import { selectStockMarket } from '../stock-market.selectors';
 import { ActionStockMarketRetrieve } from '../stock-market.actions';
+import { STOCK_MARKET_KEY } from '../stock-market.effects';
 import { State } from '../../examples.state';
+
+import { LocalStorageService } from '@app/core';
 
 @Component({
   selector: 'anms-stock-market',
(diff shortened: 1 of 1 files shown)


src/app/examples/stock-market/components/stock-market-container.component.ts
```diff
@@ -18,14 +21,23 @@ export class StockMarketContainerComponent implements OnInit, OnDestroy {
   initialized;
   stocks;
 
-  constructor(public store: Store<State>) {}
+  constructor(
+    public store: Store<State>, 
+    public localStorageService: LocalStorageService) {
+
+      this.stocks  = localStorageService.getItem(STOCK_MARKET_KEY);
```
timdeschryver: Since we're using [NgRx](https://github.com/ngrx/platform) we don't want to call services directly from our components. If we want to retrieve something it has to come from the store:

```ts
this.store
      .pipe(select(selectStockMarket), takeUntil(this.unsubscribe$))
```

I haven't taken a detailed look yet, but @tomastrajan mentioned the following in the issue:

> probably is a race condition with initializing state from local storage / reducer initial state and expecting that state in the StockMarketComponent ngOnInit method.

cloudstr21: Oh ya, i forgot to put issue number. Ok then, i will try another solutions with ngrx way, thank you

tomastrajan:
Fixed in a5c3785ed8d6a35868bc169f07e40e889087fd2e

Issue #263, opened 2018-06-26
Create new "form" example


tomastrajan:
The starter project contains lazy loaded examples module which currently contains following examples:

* todos 
* stock market
* theming
* auth only route

Create new example demonstrating "real life" complex reactive form using a form builder.

* should live in `examples/form` route
* should have dedicated component, reducer (effects if needed)

Form should have following features:

* record to be edited (eg user record with edit, save, reset button)
* use various input types (eg input, dropdown, slider, toggle)
* use unified "real life" like content (eg pizza creator, user editor, order creation or similar)
* use form validation with validation messages (eg required, email, more than ...)
* reset button ( to clear form to  original state )
...

Please feel free to suggest, enhance or improve anything!

Feel free to check out [Contributors Guide](https://github.com/tomastrajan/angular-ngrx-material-starter/blob/master/CONTRIBUTING.md) for useful tips and hints that help you get started.

timdeschryver:
Closed via #317.

PR #317, opened 2018-08-29
feat(form): Create new form example #263

Jeykairis:


tomastrajan:
Hi @Jeykairis ! Thank you very much for your contribution! There seem to be some [failing tests](https://travis-ci.org/tomastrajan/angular-ngrx-material-starter/builds/422236975?utm_source=github_status&utm_medium=notification), can you please have a look?

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
tomastrajan: Let's add this file to the `.gitignore` and not commit it.

src/app/examples/examples.module.ts
```diff
@@ -19,6 +19,16 @@ import { StockMarketService } from './stock-market/stock-market.service';
 import { ParentComponent } from './theming/parent/parent.component';
 import { ChildComponent } from './theming/child/child.component';
 import { AuthenticatedComponent } from './authenticated/authenticated.component';
+import { FormComponent } from './form/components/form.component';
+import { ReactiveFormsModule } from '@angular/forms';
+
+import {
```
tomastrajan: Let's put these things into `SharedModule`.

src/app/examples/form/components/form.component.html
```diff
@@ -0,0 +1,119 @@
+<div class="container">
+  <div class="row">
+    <div class="col">
+      <h1 class="main-heading">{{ 'anms.examples.form.title' | translate }}</h1>
+    </div>
+  </div>
+  <form [formGroup]="form">
+    <div class="row">
+      <div class="col-md-6" [ngClass]="routeAnimationsElements">
+        <h2>Contact form <mat-slide-toggle formControlName="autosave">Auto Save</mat-slide-toggle></h2>
```
tomastrajan: Let's implement all strings using the i18n (so similarly to `'anms.examples.form.title'`), it is ok to only provide english translations, it will default to english for other languages and they can be added later.

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,126 @@
...
+import { CdkTextareaAutosize } from '@angular/cdk/text-field';
+import { ROUTE_ANIMATIONS_ELEMENTS } from '@app/core';
+import { Store, select } from '@ngrx/store';
+import { Subject } from 'rxjs';
+import { State } from '@app/examples/examples.state';
+import { takeUntil } from 'rxjs/operators';
+import { selectForm } from '@app/examples/form/form.selectors';
+import { ActionFormRetrieve } from '@app/examples/form/form.actions';
+import { Form } from '@app/examples/form/form.model';
+import { MatSnackBar } from '@angular/material';
```
tomastrajan: Please, have a look into imports. I saw many unused imports in some of the changed files. Also it would be nice to follow ordering as in other files:

1. library imports (`@angular`, `@ngrx`, `rxjs`..)
2. app wide imports (`@app`, `@env`, ...)
3. module relative imports (`../form.actions`, ...)

and please add space between the groups.

 You can see it in action in todos module for example....

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,126 @@
...
+      [
+        Validators.required,
+        Validators.minLength(10),
+        Validators.maxLength(1000)
+      ]
+    ],
+    wannagift: [''],
+    age: ['', [Validators.min(18), Validators.max(99)]],
+    dob: [''],
+    address: this.fb.group({
```
tomastrajan: Is this used? I tried to fill out the form but didn't see the address fields.

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,126 @@
...
+    email: ['', [Validators.required, Validators.email]],
+    description: [
+      '',
+      [
+        Validators.required,
+        Validators.minLength(10),
+        Validators.maxLength(1000)
+      ]
+    ],
+    wannagift: [''],
```
tomastrajan: Can be please changed to something more formal like `requestGift` or `acceptsGift`, ...

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+    });
+  }
+
+  ngOnInit() {
+    this.initialized = false;
+    this.store
+      .pipe(select(selectForm), takeUntil(this.unsubscribe$))
+      .subscribe((form: any) => {
+        if (!this.initialized) {
```
tomastrajan: The whole thing works only with local storage so do we even need tracking `initialized` ?

src/app/examples/form/form.actions.ts
```diff
@@ -0,0 +1,33 @@
...
+export class ActionFormRetrieveSuccess implements Action {
+  readonly type = FormActionTypes.RETRIEVE_SUCCESS;
+
+  constructor(readonly payload: { form: any }) {}
+}
+
+export class ActionFormRetrieveError implements Action {
+  readonly type = FormActionTypes.RETRIEVE_ERROR;
+
+  constructor(readonly payload: { error: HttpErrorResponse }) {}
```
tomastrajan: As previously mentioned, this works only with local storage so do we need `HttpErrorResponse` ? On the other hand, is there some backend where we could do real `http` requests ?

src/app/examples/form/form.model.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface Form {
+  firstname: string;
+  lastname: string;
+  email: string;
+  description: string;
+  wannagift: boolean;
+  age: number;
+  dob: Date;
+  address: {
```
tomastrajan: As mentioned previously, does our form use all of these fields?

src/app/examples/stock-market/stock-market.actions.ts
```diff
@@ -4,9 +4,9 @@ import { HttpErrorResponse } from '@angular/common/http';
 import { Stock } from './stock-market.model';
 
 export enum StockMarketActionTypes {
-  RETRIEVE = '[Todos] Retrieve',
-  RETRIEVE_SUCCESS = '[Todos] Retrieve Success',
-  RETRIEVE_ERROR = '[Todos] Retrieve Error'
+  RETRIEVE = '[Stock] Retrieve',
```
tomastrajan: Great catch! Thanks!

src/assets/i18n/examples/en.json
```diff
@@ -61,6 +62,7 @@
   "anms.examples.auth.title": "Authenticated component",
   "anms.examples.auth.description1":
     "This component is only accessible for authenticated users",
+  "anms.examples.form.title": "Complex reactive form using a form builder",
```
tomastrajan: Please add also additional strings to be used with `i18n`, thank you!

src/app/examples/form/components/form.component.ts
Jeykairis: Yes this is. Fields should appears after the dob is filled in.

src/app/examples/form/components/form.component.ts
Jeykairis: It's used to restore the state of the form from the local storage.
Do you have a better way to handle it?

By the way, this is only working while navigating but not when the app is relaunched.
That looks like something related to #315.

src/app/examples/form/form.model.ts
Jeykairis: Yes it does. But the model is not really used as it should be to replace `any` in the effect.

Do you see a way to improve the loop getting data from the local storage using the `Form` type?
Then what is the cleanest way to map the `FormGroup` to a `Form` before dispatching it?

src/app/examples/stock-market/stock-market.actions.ts
Jeykairis: Thank you!

tomastrajan:
Hi @Jeykairis ! Thank you for all the changes! There is still one issue when I edit age by writing the number instead of clicking on the small arrows of the number input the application seems to enter state of a infinite loop and gets stuck. The ngrx actions are dispatched in a loop which doesn't stop and crashes the browser tab... 

I was testing it now it also happens when I use number input stepper controls (up / down arrows) it starts infinite loop and crashes the tab ://

tomastrajan:
Also lets chat a bit in general about the actions and effects. We have a form and we need to:

1. initialize form in the beginning if some state was saved in the local storage - this can happen automatically just by using `this.store.pipe(select(selectForm), takeUntil(this.unsubscribe$))` in combination with the `export const FORM_KEY = 'EXAMPLES.FORM';`. This will initialize `Store` state in `examples.form` at the application startup automatically so we don't need to do it.
2. we need to update store form data on form update so we need something like `ActionFormUpdate` so that we set the form values into the store
3. we need to save current form into local storage if person clicked `save` or has `autosave` enabled, this means we need `ActionFormSave` action to store it into localstorage using `ngrx` effects.

Please have a look on `Todos` implementation which is a very similar, it's basically a very simple form too.

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,121 @@
...
+  countries = ['Belgium', 'Switzerland'];
+
+  constructor(
+    private fb: FormBuilder,
+    private store: Store<State>,
+    public snackBar: MatSnackBar
+  ) {
+    this.form.valueChanges.subscribe((f: FormGroup) => {
+      const today = new Date();
+      this.minDate = new Date(
```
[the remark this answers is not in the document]
Jeykairis: Unfortunately I cannot reproduce the delay when filling in the age field.
This is working properly on my side and the `minDate` is a constraint on the date picker which does not trigger the `form.valueChanges` for me...

Jeykairis:
Thanks a lot for your guidance!

tomastrajan:
Hi @Jeykairis ! I am reviewing and reworking your PR local as it still has some issues and it proved to be more complex than originally thought. The main issues are around age and related min / max date validation which often triggers infinite loop and crashes the browser tab... I will let you know when I come to some solution and maybe I will commit it to your fork so we can then review it and merge together...

Anyway, most of your commit can and will be used so thanks!

tomastrajan:
Hi @Jeykairis ! I cloned your PR locally and was playing around and it still has some problems we will have to figure out. On the other hand I would love to merge your contribution which contains a lot of very useful stuff. Could we do it in such a way that you will simplify the form into lets say 4 inputs with name, surname, birthday (datepicker), requestGift (boolean). So that we have a really working example and then we can build on top of that? cc: @timdeschryver

Jeykairis:
Hi @tomastrajan, thanks for reviewing it and trying to figure out what is wrong.
Indeed, I tried to add more complex use case to change the form following previous values but I also remarked errors like `ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'ngClass: col-4'. Current value: 'ngClass: col'.`.
So, I simplified the form to keep controls already implemented avoiding errors. Feel free to improve and update it at your best convenience.
Nevertheless, it would be great to show the cleanest way to update the form UI and constraints while filling in it.

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  constructor(
+    private fb: FormBuilder,
+    private store: Store<State>,
+    private translate: TranslateService,
+    public snackBar: MatSnackBar
+  ) {}
+
+  ngOnInit() {
+    this.store
+      .pipe(select(selectForm), takeUntil(this.unsubscribe$))
```
timdeschryver: Would it make sense to only take the first emitted value, and fill in our form with it? 
This would mean that we don't need `ActionFormSave` anymore, we could use `ActionFormUpdate` to persist to local storage.

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,100 @@
...
+      .subscribe(form => {
+        if (!this.form.dirty) {
+          this.form.patchValue(form.form);
+        }
+        this.store.dispatch(new ActionFormSave({ form: form }));
+      });
+
+    this.form.valueChanges.subscribe((f: FormGroup) => {
+      if (f['autosave']) {
```
timdeschryver: Use the `filter` operator, instead of the inline filter.
Could you also rename `f` to `form` please?

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  save() {
+    this.store.dispatch(new ActionFormUpdate({ form: this.form.value }));
+  }
+
+  reset() {
+    this.form.reset();
+    this.form.clearValidators();
+    this.form.clearAsyncValidators();
+    this.store.dispatch(new ActionFormUpdate({ form: {} }));
```
timdeschryver: I think it would be cleaner if we would use a reset action instead of using `ActionFormUpdate` for it.

src/app/examples/form/form.actions.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor(readonly payload: { form: FormState }) {}
+}
+
+export class ActionFormUpdate implements Action {
+  readonly type = FormActionTypes.UPDATE;
+
+  constructor(readonly payload: { form: any }) {}
+}
+
+export class ActionFormRetrieve implements Action {
```
timdeschryver: Can be deleted, isn't used.

src/app/examples/form/form.actions.ts
```diff
@@ -0,0 +1,32 @@
+import { Action } from '@ngrx/store';
+import { Observable } from 'rxjs';
+
+import { Form, FormState } from './form.model';
+
+export enum FormActionTypes {
+  SAVE = '[Form] Save',
+  UPDATE = '[Form] Update',
+  RETRIEVE = '[Form] Retrieve'
```
timdeschryver: Can be deleted, isn't used.

src/app/examples/form/components/form.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    this.store
+      .pipe(select(selectForm), takeUntil(this.unsubscribe$))
+      .subscribe(form => {
+        if (!this.form.dirty) {
+          this.form.patchValue(form.form);
+        }
+        this.store.dispatch(new ActionFormSave({ form: form }));
+      });
+
+    this.form.valueChanges.subscribe((f: FormGroup) => {
```
timdeschryver: I think it would make sense to debounce the user input.
What do you think?

timdeschryver: Wouldn't the type of `f` be `Form` here?

src/app/examples/form/components/form.component.html
```diff
@@ -0,0 +1,81 @@
...
+    </div>
+  </div>
+  <form [formGroup]="form" (ngSubmit)="onSubmit()">
+    <div class="row justify-content-center">
+      <div class="col-md-6" [ngClass]="routeAnimationsElements">
+        <mat-slide-toggle formControlName="autosave">{{ 'anms.examples.form.autosave' | translate }}</mat-slide-toggle>
+        <h2>{{ 'anms.examples.form.subtitle1' | translate }}</h2>
+        <div class="row">
+          <mat-form-field class="col" [ngClass]="routeAnimationsElements">
+            <input matInput placeholder="{{ 'anms.examples.form.placeholder1' | translate }}" formControlName="username" required>
```
timdeschryver: Is there a reason to put `required` in the inputs?
Because we're using reactive forms, we don't need to I think but I might be missing something

src/app/examples/form/form.reducer.ts
```diff
@@ -15,6 +15,11 @@ export function formReducer(
         ...state,
         form: action.payload.form
       };
+    case FormActionTypes.RESET:
+      return {
```
timdeschryver: We can just return `initialState` here.

src/app/examples/form/form.actions.ts
```diff
@@ -1,32 +1,21 @@
...
+  constructor(readonly payload: { form: Form }) {}
 }
 
-export class ActionFormRetrieve implements Action {
-  readonly type = FormActionTypes.RETRIEVE;
+export class ActionFormReset implements Action {
+  readonly type = FormActionTypes.RESET;
 
-  constructor(readonly payload: { form: Form }) {}
+  constructor() {}
```
timdeschryver: remove the empty ctor

Jeykairis:
Yes I run it and it lists some files without any formatting to do.
```txt 
> angular-ngrx-material-starter@6.9.0 prettier /Users/jeremykairis/_Alab/github/contribution/angular-ngrx-material-starter
> prettier {src,e2e}/**/*.{ts,json,md,scss} --write

src/app/app-routing.module.ts 302ms
src/app/app.component.spec.ts 14ms
src/app/app.component.ts 92ms
src/app/app.module.ts 10ms
src/environments/environment.prod.ts 10ms
src/environments/environment.test.ts 9ms
src/environments/environment.ts 13ms
src/testing/utils.ts 25ms
src/assets/i18n/de.json 56ms
src/assets/i18n/en.json 14ms
src/assets/i18n/es.json 15ms
src/assets/i18n/examples/de.json 9ms
src/assets/i18n/examples/en.json 14ms
src/assets/i18n/examples/es.json 10ms
src/assets/i18n/examples/fr.json 19ms
src/assets/i18n/examples/pt-br.json 9ms
src/assets/i18n/examples/sk.json 7ms
src/assets/i18n/fr.json 12ms
src/assets/i18n/pt-br.json 21ms
src/assets/i18n/sk.json 15ms
src/tsconfig.app.json 6ms
src/tsconfig.spec.json 4ms
src/app/app.component.scss 140ms
src/app/app.component.scss-theme.scss 14ms
src/themes/black-theme.scss 3ms
src/themes/default-theme.scss 5ms
src/themes/light-theme.scss 4ms
src/themes/nature-theme.scss 4ms
e2e/src/app.e2e-spec.ts 13ms
e2e/src/app.po.ts 7ms
e2e/tsconfig.e2e.json 2ms```

Do you know what happens ?

**Comment by tomastrajan** [comment]
@Jeykairis seems to be issue with 
```
> prettier {src,e2e}/**/*.{ts,json,md,scss} --list-different
src/app/examples/examples.module.ts
```

As in [build log](https://travis-ci.org/tomastrajan/angular-ngrx-material-starter/builds/428963495#L541).

**Comment by tomastrajan** [comment]
@Jeykairis maybe it's because of different environment, the glob `/**/*` seems to only get expanded for one level of directories but not in the sub-levels therefore skipping these files, what are you using ?

**Comment by Jeykairis** [comment]
@tomastrajan Thank you for the highlight!

File: src/app/examples/form/form.model.ts [hunk 22]
```diff
@@ -0,0 +1,14 @@
+export interface Form {
+  username: string;
+  password: string;
+  email: string;
+  description: string;
+  requestGift: boolean;
+  age: number;
```
timdeschryver: Could it be that this is a leftover from the rebase?
I think this can be removed?

src/app/settings/settings.model.ts
```diff
@@ -2,7 +2,7 @@ import { AppState } from '@app/core';
 
 export const NIGHT_MODE_THEME = 'BLACK-THEME';
 
-export type Language = 'en' | 'sk' | 'de' | 'fr' | 'es' | 'pt-br';
+export type Language = 'en' | 'sk' | 'de' | 'fr' | 'es';
```
timdeschryver: Could it be that this is a leftover from the rebase, or was it intentionally to remove `pt-br`?

src/app/settings/components/settings-container.component.ts
```diff
@@ -35,8 +35,7 @@ export class SettingsContainerComponent implements OnInit, OnDestroy {
     { value: 'de', label: 'de' },
     { value: 'sk', label: 'sk' },
     { value: 'fr', label: 'fr' },
-    { value: 'es', label: 'es' },
-    { value: 'pt-br', label: 'pt-br' }
+    { value: 'es', label: 'es' }
```
timdeschryver: Could it be that this is a leftover from the rebase, or was it intentionally to remove `pt-br`?

Jeykairis:
Yes indeed, thank you @timdeschryver !

tomastrajan:
This will be the most epic PR in the repo history, very happy to see it is getting into good shape, thanks guys!

Jeykairis:
@timdeschryver Thanks for your attention, I missed the comment. 
@tomastrajan Thanks for welcoming me on this awesome starter!

Thank you both for your time and feedbacks.
I hope we will be able to improve this form to cover corner cases as initially intended.

tomastrajan:
Hmm I have already saw this [build error](https://travis-ci.org/tomastrajan/angular-ngrx-material-starter/builds/430248694#L610) before when adding support for `camelCase` state properties to be stored in local storage. I didn't really solve it, just by removing a one test which was not about the specific test just **ANY** one test. I couldn't really figure out what was wrong. Commenting out any of the passing test in other file seemed to fix the build ://

timdeschryver:
Weird... this shouldn't be blocking the PR in my opinion, I think this can be merged.

tomastrajan:
@timdeschryver agreed!

PR #323, opened 2018-09-05
(fix) : Disable page animation in Safari fixes #322

### LINKED ISSUE #322: The animation `routeAnimationsElements` looks bad on safari.
Just launch the app on safari to reproduce the issue : 
routeanimationselements_safari

Jeykairis:


tomastrajan:
@Jeykairis thank you very much for the contribution! Just to be sure, id did help, it worked for you on Safari, right?

tomastrajan:
@Jeykairis thanks again! Great catch, we want to make it work for all people / browsers so this is very appreciated!

Issue #286, opened 2018-07-20
Create new example that implements CRUD functionality using @ngrx/entity


tomastrajan:
Starter project already contains multiple example implementations ( todos, stock market, theming of nested components, ... ),. It would be great to add another example wich implements CRUD (create, read, update, delete) functionality for a collection of entities using `@ngrx/entity`.

It would make sense to implement it by following [official introductory blog post](https://medium.com/ngrx/introducing-ngrx-entity-598176456e15).

Example should be created in examples module in a separate directory, we can call it `crud` ( same for the menu name).

The implementation will need (similar to todos):

- component ( implementation, template, styles, tests )
- reducer, tests

As for the entities, it can be almost anything you want. Books, movies, users, anything roughly normal.
Couple of properties: `id`, `name`, `author`, ...

Layout can be a list of entities on the left and entity detail editor on the right which loads entity to edit when user clicks on a entity in list.

This issue was inspired by request in #285

tomastrajan:
Closed by https://github.com/tomastrajan/angular-ngrx-material-starter/commit/460a94c2c61a0511942946c0781d6da8c8bc9e8c

PR #325, opened 2018-09-09
feat(crud) : New example of CRUD functionality using @ngrx/entity #286

Jeykairis:


tomastrajan:
Hi @Jeykairis, this is very cool thank you!

src/app/examples/crud/books.model.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface Book {
+  id: string;
+  title: string;
+  author: string;
+  description: string;
+}
+
+export interface BookState extends EntityState<Book> {
+  selectedBookId: string | number | null;
```
timdeschryver: Remove the `number` type.

src/app/examples/crud/books.reducer.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  it('should update a book', () => {
+    const id = TEST_INITIAL_STATE.ids[0].toString();
+    const action = new UpdateOne(id, {
+      title: 'updated',
+      author: 'updated',
+      description: 'updated'
+    });
+    const state = bookReducer(TEST_INITIAL_STATE, action);
+    expect(state.entities[id].title).toEqual('updated');
```
timdeschryver: Personally I would like it if this would be written in one expect instead of three.

src/app/examples/crud/books.reducer.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      description: 'test'
+    });
+    const state = bookReducer(TEST_INITIAL_STATE, action);
+
+    expect(state.ids.length).toEqual(2);
+    expect(state.entities['1234'].title).toEqual('test');
+  });
+
+  it('should update a book', () => {
+    const id = TEST_INITIAL_STATE.ids[0].toString();
```
timdeschryver: Why is the `toString` needed here?

src/app/examples/crud/books.reducer.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      description: 'updated'
+    });
+    const state = bookReducer(TEST_INITIAL_STATE, action);
+    expect(state.entities[id].title).toEqual('updated');
+    expect(state.entities[id].author).toEqual('updated');
+    expect(state.entities[id].description).toEqual('updated');
+  });
+
+  it('should remove a book', () => {
+    const id = TEST_INITIAL_STATE.ids[0].toString();
```
timdeschryver: Why is the `toString` needed here?

src/app/examples/crud/books.reducer.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    const action = new DeleteOne(id);
+    const state = bookReducer(TEST_INITIAL_STATE, action);
+    expect(state.entities[id]).toBe(undefined);
+  });
+
+  it('should select a book', () => {
+    const id = TEST_INITIAL_STATE.ids[0].toString();
```
timdeschryver: Why is the `toString` needed here?

src/app/examples/crud/books.selectors.ts
```diff
@@ -0,0 +1,31 @@
...
+  (state: ExamplesState) => state.books
+);
+
+export const getSelectedBookId = (state: BookState) => state.selectedBookId;
+export const selectCurrentBookId = createSelector(
+  selectBooks,
+  getSelectedBookId
+);
+
+export const selectBooksIds = createSelector(selectBooks, selectIds);
```
timdeschryver: This can be removed, it isn't used.

src/app/examples/crud/books.selectors.ts
```diff
@@ -0,0 +1,31 @@
...
+export const selectBooksEntities = createSelector(selectBooks, selectEntities);
+export const selectAllBooks = createSelector(selectBooks, selectAll);
+export const selectBooksTotal = createSelector(selectBooks, selectTotal);
```
timdeschryver: This can be removed, it isn't used.

src/app/examples/crud/components/crud.component.ts
```diff
@@ -0,0 +1,105 @@
...
+    this.store
+      .pipe(select(selectBooks), takeUntil(this.unsubscribe$))
+      .subscribe(books => {
+        this.store.dispatch(new ActionBooksPersist({ books }));
+      });
+    this.store
+      .pipe(
+        select(selectAllBooks),
+        takeUntil(this.unsubscribe$),
+        map(b => b.sort((b1, b2) => (b1.title > b2.title ? 1 : -1)))
```
timdeschryver: Would it make sense to this when we create the adapter? (it has a sort property)

src/app/examples/crud/components/crud.component.html
```diff
@@ -0,0 +1,80 @@
...
+          <button type="button" mat-raised-button [ngClass]="routeAnimationsElements" (click)="clearForm(bookForm)"
+            color="accent">{{'anms.examples.crud.new' | translate }}
+          </button>
+          <button mat-raised-button color="primary" [ngClass]="routeAnimationsElements">{{
+            'anms.examples.crud.save' | translate }}
+          </button>
+        </div>
+      </form>
+      <div [ngClass]="routeAnimationsElements" [style.display]="isEditing ? 'none' : 'inherit'">
+        <h3>{{selectedBook?.title}}</h3>
```
timdeschryver: Perhaps it would be better to use an `*ngIf` on the div? Instead of the `?.` operator

src/app/examples/crud/components/crud.component.ts
```diff
@@ -0,0 +1,105 @@
...
+  };
+  bookFormGroup = this.fb.group(this.emptyBook);
+  books: Book[];
+  selectedBook: Book | null;
+  isEditing: boolean;
+
+  constructor(public store: Store<State>, public fb: FormBuilder) {}
+
+  ngOnInit() {
+    this.store
```
timdeschryver: Would it make sense to handle the persistence inside effects, instead of the component itself?

src/app/examples/crud/components/crud.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        map(b => b.sort((b1, b2) => (b1.title > b2.title ? 1 : -1)))
+      )
+      .subscribe(books => {
+        this.books = books;
+      });
+    this.store
+      .pipe(select(selectCurrentBook), takeUntil(this.unsubscribe$))
+      .subscribe(book => {
+        if (book !== undefined) {
+          this.selectedBook = book;
```
timdeschryver: I don't like it when we subscribe to selectors and assign the values like this.
I prefer using the `async` pipe, we can do this by creating a dumb component.
Another option would be to use asign the value in combination with the `async` pipe, `*ngIf="observable | async as obs"`.
Thoughts?

src/app/examples/crud/components/crud.component.ts
```diff
@@ -0,0 +1,105 @@
...
+  selector: 'anms-crud',
+  templateUrl: './crud.component.html',
+  styleUrls: ['./crud.component.scss']
+})
+export class CrudComponent implements OnInit {
+  private unsubscribe$: Subject<void> = new Subject<void>();
+
+  routeAnimationsElements = ROUTE_ANIMATIONS_ELEMENTS;
+
+  emptyBook: Book = {
```
timdeschryver: I think it will be cleaner if we would create a factory function to create an empty book.

src/app/examples/crud/components/crud.component.ts
Jeykairis: Yes it could be cleaner but i'm searching a way to achieve it. 
I'm trying something with the effect bellow but without success.
```typescript
@Effect({dispatch: false})
  persistBooks() {
    return this.actions$.pipe(
      ofType<AddOne|UpdateOne|DeleteOne>(
        BookActionTypes.ADD_ONE,
        BookActionTypes.UPDATE_ONE,
        BookActionTypes.DELETE_ONE),
      tap(_ => {
        this.store.select(state => {
          return this.localStorageService.setItem(
            BOOKS_KEY,
            state.examples.books
          );
        });
      })
    );
```
@timdeschryver or @tomastrajan do you have an idea?

timdeschryver: I think instead of using `select` in the `tap` operator, you could use `withLatestFrom(this.store.pipe(select(selector)))`.

If this isn't the case I suppose we can solve it in a separate issue, because I think this is the way how it's currently done in every example.

timdeschryver: see the following article, find `4. Using a selector inside your effects`:

https://blog.angularindepth.com/start-using-ngrx-effects-for-this-e0b2bd9da165

timdeschryver: OK thanks for trying 😄 , let's do this in an other issue and in every example.

timdeschryver:
LGTM, thanks @Jeykairis !

tomastrajan:
Thank you @Jeykairis and @timdeschryver great stuff!!

tomastrajan:
BTW @timdeschryver, the article is amazing and we should definitely move in direction of using effects for stuff like notifications 👍 Also, any ideas on this ?

```
 @Effect()
  something() {
    return this.actions$.pipe(/*...*/);
  }
```
vs
```
  @Effect()
  something = this.actions$.pipe(/*...*/);
```

Does it behave exactly the same or is there some difference? Cuz if not then the second one is shorter :)

tomastrajan:
@timdeschryver I see, makes sense, thanks!

PR #327, opened 2018-09-13
fix(todos): snackbar not being translated

### LINKED ISSUE #324: Snackbar is not being translated
1. Filter todos
2. Switch language to not english
3. FIlter todos again

Notice that the snackbar is not translated.

aideslucas:
added translation for all snackbar messages,
on filter, toggle, added, and removed

fixes #324


tomastrajan:
Hi @aideslucas !

Thank you very much for your contribution! Could you please use `this.translateService.instant(key)` instead of aync `.get()` method? It will simplify implementation significantly and based on our routing / lazy loaded module structure we can be sure that the translations are already present when we need to translate these notifications. Thank you!

aideslucas:
Hey @tomastrajan ,
switched the functions, first time using ngx-translante didnt see the instant method in the docs

PR #341, opened 2018-09-19
fix(ui): Cosmetic changes to improve responsive translated views

Jeykairis:
## What:

Some cosmetic changes to make the starter looks better with every languages on every screen sizes.

Issue number: N/A

### DIFF
--- a/src/app/examples/examples/examples.component.scss
+++ b/src/app/examples/examples/examples.component.scss
@@ -3,6 +3,7 @@
 nav {
   margin-bottom: 20px;
   position: relative;
+  overflow: scroll;
 
   .mat-tab-link {
     min-width: 130px;
--- a/src/app/examples/examples/examples.component.scss
+++ b/src/app/examples/examples/examples.component.scss
@@ -2,17 +2,15 @@
 
 nav {
   margin-bottom: 20px;
-  position: relative;
-  overflow: scroll;
 
   .mat-tab-link {
-    min-width: 130px;
+    min-width: 120px;
+    padding: 0 15px;
   }
 
-  @media (max-width: map-get($grid-breakpoints, sm)) {
+  @media (max-width: map-get($grid-breakpoints, md)) {
     .mat-tab-link {
       min-width: 0;
-      padding: 0 15px;
     }
   }
 }
--- a/src/index.html
+++ b/src/index.html
@@ -25,11 +25,6 @@
       height: 100%;
     }
 
-    .app-loading img {
-      width: 100px;
-      height: 100px;
-    }
-
     .app-loading .spinner {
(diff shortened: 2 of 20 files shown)


src/app/examples/examples/examples.component.scss
```diff
@@ -3,6 +3,7 @@
 nav {
   margin-bottom: 20px;
   position: relative;
+  overflow: scroll;
```
tomastrajan: This seems to be unfortunately  broken in Google Chrome...
anms-error

tomastrajan:
@Jeykairis also what could be worth investigating is to make that sub-menu responsive in a sense that on small screens it will be replaced by the dropdown (select) element, thoughts ?

src/index.html
```diff
@@ -71,7 +76,7 @@
 <body>
   <anms-root>
     <div class="app-loading">
-      <div class="logo"></div>
+      <img src="./assets/logo.png">
```
tomastrajan: This will not work for the prod build, that was the reason to extract this to css because that way it is processed by the Weback as an asset because the `styles-app-loading.css` is referenced in the `angular.json` as a external stylesheet. Please revert this change.

tomastrajan: @Jeykairis hmm, looks like this was broken when migrating to Angular CLI 6+ with new `angular.json` format because as it seems the `styles-app-loading.scss` is not referenced anywhere in the whole application... The correct way would be to add it to the `styles` in the `angualr.json`. Another thing worth trying would be to just add the `logo.png` into the `assets` in `angular.json` but then it is the question what would happen during the prod build (eg you don't get the content hashing and cache busting) so maybe it is still better to stick to the `styles-app-loading.scss` which generates hashed asset (eg `logo-24rfsdgd23r.png`).

Jeykairis:
Haha, it was looking good for chrome on a mac but it was a dirty fix. I cleaned it and implemented the select for smallest screens.
By the way, we should take care of the material specs regarding the sub-nav https://material.io/design/components/tabs.html#behavior.

Thanks @timdeschryver for fixing the autosave !

src/index.html
Jeykairis: Okay, thank you for the explanations!
I tried to import the `styles-app-loading.scss` into the `styles.scss` and the logo appears but not the spinner...
I try to add the `styles-app-loading.scss` directly in the `angular.json`.

tomastrajan: @Jeykairis the way this behaves is that the spinner should be shown immediately because it is the part of initial `index.html` template, after that when the styles bundle is loaded for `styles-app-loading.scss` the logo should appear and after that when the whole application is loaded it will replace the spinner and the logo with the app itself ;)

PR #351, opened 2018-09-21
Fix memory leak in crud component

simply10w:
## What:

Emit on destroy to close all open subscriptions in crud component.

Issue number: N/A


timdeschryver:
Great, thanks @simply10w !

Issue #266, opened 2018-06-28
Create unit tests for stock market reducer


tomastrajan:
Create tests for `stock-market.reducer.ts` in `stock-market.reducer.spec.ts` file either by following [official ngrx testing guide](https://github.com/ngrx/platform/blob/master/docs/store/testing.md) or feel free to suggest any improvements.

Feel free to check out [Contributors Guide](https://github.com/tomastrajan/angular-ngrx-material-starter/blob/master/CONTRIBUTING.md) for useful tips and hints that help you get started.

timdeschryver:
Closed via #353.

PR #353, opened 2018-09-22
test(examples): create unit tests for stock market reducer #266

ZackDeRose:
## What:

Unit tests for `stock-market.reducer`.

Issue number: #266

### DIFF
--- a/src/app/examples/stock-market/stock-market.reducer.spec.ts
+++ b/src/app/examples/stock-market/stock-market.reducer.spec.ts
@@ -1,0 +1,93 @@
+import { Stock, StockMarketState } from './stock-market.model';
+import { HttpErrorResponse } from '@angular/common/http';
+import { stockMarketReducer, initialState } from './stock-market.reducer';
+import {
+  StockMarketActions,
+  ActionStockMarketRetrieve,
+  ActionStockMarketRetrieveError,
+  ActionStockMarketRetrieveSuccess
+} from './stock-market.actions';
+
+const originalState: StockMarketState = {
+  symbol: 'LCLT',
+  loading: false,
+  stock: {
+    symbol: 'LCLT',
+    exchange: 'TestExchange',
+    last: '1.00',
+    ccy: 'USD',
+    change: '+0.05',
+    changePositive: true,
+    changeNegative: false,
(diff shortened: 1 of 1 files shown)


src/app/examples/stock-market/stock-market.reducer.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      const stock = {
+        symbol: 'AEONS',
+        exchange: 'TestExchange2',
+        last: '2.00',
+        ccy: 'USD',
+        change: '+0.10',
+        changePositive: true,
+        changeNegative: false,
+        changePercent: '+5%'
+      } as Stock;
```
tomastrajan: Does this have to be `as Stock` or could it also be `const stock: Stock = {}` because it has all the properties?

src/app/examples/stock-market/stock-market.reducer.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    symbol: 'LCLT',
+    exchange: 'TestExchange',
+    last: '1.00',
+    ccy: 'USD',
+    change: '+0.05',
+    changePositive: true,
+    changeNegative: false,
+    changePercent: '+5%'
+  }
+} as StockMarketState;
```
tomastrajan: Could this be also `const originalState: StockMarketState = {}` because it has all the needed properties?

ZackDeRose: I think they are essentially the same, but I agree that your suggestion reads better/the expected way.

tomastrajan:
@ZackDeRose thank you very much for the adjustments, the last hing would be if you could rebase two commits into one. You could use `git rebase -i HEAD^2` and then change `pick` to `f` for the second commit. After that you could `git push --force` to your branch. Check out [this tutorial](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) for more detail (just go for `f` / `fixup` instead of `edit`), thank you!

ZackDeRose:
Got it! Thanks for the tutorial link. Usually end up using Squash & Merge on Github for these.

PR #354, opened 2018-09-22
Perf/on push

### LINKED ISSUE #352: Using OnPush to improve performance
## Describe the feature you'd like:

Adding OnPush change detection to all components. 

Since we already are utilizing @ngrx/store with rxjs streams and async pipe, we get markForCheck for free in most situations.

By adding on push, only some parts of the application would need to be checked when change detection runs. Currently it runs over the entire tree.

That way we would also enforce good practices from the get go, avoiding mutating state, and if mutation needs to happen, triggering change detection manually.

[ x] Yes (Assistance is provided if you need help submitting a pull request)  
[ ] No

simply10w:
## What:

Added OnPush change detection strategy to all components.

Updated `angular.json` to generate components with OnPush by default.

Closes #352

Issue number: #352

### DIFF
--- a/src/app/examples/crud/components/crud.component.ts
+++ b/src/app/examples/crud/components/crud.component.ts
@@ -1,5 +1,11 @@
 import { v4 as uuid } from 'uuid';
-import { Component, OnInit, OnDestroy } from '@angular/core';
+import {
+  Component,
+  OnInit,
+  OnDestroy,
+  ChangeDetectionStrategy,
+  ChangeDetectorRef
+} from '@angular/core';
 import { FormBuilder, NgForm } from '@angular/forms';
 import { Router } from '@angular/router';
 import { takeUntil } from 'rxjs/operators';
@@ -16,7 +22,8 @@ import { selectSelectedBook, selectAllBooks } from '../books.selectors';
 @Component({
   selector: 'anms-crud',
   templateUrl: './crud.component.html',
-  styleUrls: ['./crud.component.scss']
+  styleUrls: ['./crud.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class CrudComponent implements OnInit, OnDestroy {
   private unsubscribe$: Subject<void> = new Subject<void>();
@@ -39,7 +46,8 @@ export class CrudComponent implements OnInit, OnDestroy {
   constructor(
     public store: Store<State>,
     public fb: FormBuilder,
-    private router: Router
+    private router: Router,
+    private cd: ChangeDetectorRef
   ) {}
 
   ngOnInit() {
@@ -49,7 +57,10 @@ export class CrudComponent implements OnInit, OnDestroy {
         select(selectSelectedBook),
         takeUntil(this.unsubscribe$)
       )
-      .subscribe(book => (this.selectedBook = book));
+      .subscribe(book => {
+        this.selectedBook = book;
+        this.cd.markForCheck();
+      });
   }
 
   ngOnDestroy() {
--- a/src/app/examples/crud/components/crud.component.ts
+++ b/src/app/examples/crud/components/crud.component.ts
@@ -1,15 +1,8 @@
 import { v4 as uuid } from 'uuid';
-import {
-  Component,
-  OnInit,
-  OnDestroy,
-  ChangeDetectionStrategy,
-  ChangeDetectorRef
-} from '@angular/core';
+import { Component, ChangeDetectionStrategy } from '@angular/core';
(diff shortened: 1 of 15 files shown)


src/app/examples/crud/components/crud.component.ts
```diff
@@ -40,13 +47,20 @@ export class CrudComponent implements OnInit, OnDestroy {
...
+  ) {}
 
   ngOnInit() {
     this.books$ = this.store.pipe(select(selectAllBooks));
     this.store
       .pipe(select(selectSelectedBook), takeUntil(this.unsubscribe$))
-      .subscribe(book => (this.selectedBook = book));
+      .subscribe(book => {
+        this.selectedBook = book;
+        this.cd.markForCheck();
```
tomastrajan: Lets rework this component to **NOT** subscribe to this in the `ngOnInit` but in the template using `*ngIf=selectedBook$ | async as selectedBook` so that we don't have to manually trigger `markForCheck`. Let's do this for all cases when this is needed.

Another side effec could be that many of these components will no longer need `OnDestroy` and  `private unsubscribe$: Subject<void> = new Subject<void>();`...

simply10w: Deal 👍  

I didn't want to refactor a lot without approval, but you are right in the main comment, this is the point. With that:

1. we'll have a more declarative syntax,
2. remove manual unsubscriptions (automatically we remove possibility for human error having a memory leak if forgot to unsubscribe) 
3. make in general components easier to reason about because we don't need to keep local state (This also comes with the benefit that it prevents developers accidentally mutating the data got from the store, since they will not operate on it, components just get what they need and pass it down).

tomastrajan: @simply10w cool, the points make sense. I would like to add that there is one disadvantage and that is that some of the **local** state is now kept in the template, but I have to agree it's usualy pretty minimal. If we follow naming convention like `something$` is then unwrapped in template to `something` it should be easy to navigate and reason about.

Anyway looking forward to the final PR, thank you again!

tomastrajan: @simply10w  I am in favor of the second example where we pass `bookId` from the template.

tomastrajan: @simply10w btw thanks for a very nice communication of the proposals, very appreciated!

simply10w: @tomastrajan You're welcome, i'm really glad to be a part of this and help out.

timdeschryver: Agreeing with everything that is said here.
The only thing I want to add it, if we find that the HTML will become to bloated/big we can always split it up into multiple components.

simply10w: Splitting up might be a good idea even for performance reasons, for example in `todos-container` the todo list is in the same template with the `big-input`, every event that happens on big-input will trigger change detection over the entire list and each item has a lot of bindings(mostly translation pipes).

But if we encapsulate the list in a component with OnPush, if input has not changed it won't need to check the list at all.

tomastrajan: @simply10w yeah that's true, there is a lot of room for breaking stuff up into multiple components. I think that's also one of the killer features of Angular that it is possible to come up with working solution in one "big template" which then can be broken up further into more discrete components.

simply10w: @tomastrajan Yea, it gives you a lot of flexibility to quickly prototype and later when you can more clearly see correct abstractions, pull them out with ease.

tomastrajan:
@simply10w just wanted to ask if everything is all right, there is no hurry with this anyway but can we be of help?

simply10w:
@tomastrajan  Hey, yea all is good, just was a really busy week at work, so didn't have time to jump on this, will wrap it up during next week if it is not a problem? sorry for the delay

tomastrajan:
@timdeschryver this got pretty stale but would be great to have it, do we create a new issue?

timdeschryver:
I think this can be merged after a quick rebase? We can indeed fine tune in a new issue.

simply10w:
@tomastrajan  @timdeschryver  

I've refactored stream handling to be more declarative using streams and async pipe. Imperative mark for check call has been removed from all places expect from the `SettingsContainerComponent` because current implementation much more simplifies both the template and the component logic.

Sorry guys for delaying this, was swamped on life/work front, but i've managed to grab an hour to finalize this.

timdeschryver:
Hey, thanks for the commit @simply10w !
I'll take a look later today.

Issue #331, opened 2018-09-18
Use @ngrx-router-store


Jeykairis:
It would be great to show how to use @ngrx-router-store.

I'm trying to figure out how to achieve it with the CustomSerializer from the ngrx doc.
The goal could be to get the selected book from the url managing errors.

Do you have other ideas to show advanced use cases or any starting point?

timdeschryver:
Great idea, I'm OK with it. 

I'm not entirely sure what you mean with
> The goal could be to get the selected book from the url managing errors

Also with NgRx 6.1 it isn't needed anymore to provide a serialized, this comes baked in with NgRx itself. If needed are still able to provide a custom serializer tho.

Jeykairis:
After investigation I found https://github.com/ngrx/platform/issues/1262#issuecomment-421066044 and https://github.com/ngrx/platform/commit/e0059bfca6d130061ae6af8f4fa27e97552b60de .

It looks like the doc was updated but the default serializer is not yet available and I don't find the way to connect the custom one. When should it be released?

Otherwise, by managing errors I wanted to say handle the case where the id of the book passed by url is not available in the store through selectors. What else could be done with the router-store?

timdeschryver:
The link you provided, adds more router actions to ngrx-store. 
I was referring to https://github.com/ngrx/platform/commit/389cd78f3517a7ff32930dc1d8187aa7fca712f8 and I thought this would be available in 6.1 but I might be wrong.

> Otherwise, by managing errors I wanted to say handle the case where the id of the book passed by url is not available in the store through selectors. What else could be done with the router-store?

This shouldn't be handled by selectors. A selector is a pure function meaning it should just return an output based on the input. In other words it should to navigate to another page, if this is what you meant.

But we could indeed pluck the book id from the url, and show the selected book.
I briefly touch on this in [NgRx: Parameterized selectors](https://blog.angularindepth.com/ngrx-parameterized-selector-e3f610529f8) and [Sharing data between modules is peanuts.](https://blog.angularindepth.com/sharing-data-between-modules-is-peanuts-3f40fcb38dd) if you need an example. And there was also an excellent talk by Todd Motto and David East during ng-conf [NgRx Selectors How to stop worrying about your Store structure](https://youtu.be/Y4McLi9scfc?t=15m17s)

Jeykairis:
Thank you for the explanations and great articles!
[...]

PR #355, opened 2018-09-23
Use @ngrx-router-store #331

Jeykairis:
## What:

The goal is to use ngrx-router-store to get the selected book from the url.

Issue number: 331

### DIFF
--- a/src/app/core/core.state.ts
+++ b/src/app/core/core.state.ts
@@ -3,6 +3,7 @@ import {
   MetaReducer,
   createFeatureSelector
 } from '@ngrx/store';
+import { routerReducer, RouterReducerState } from '@ngrx/router-store';
 import { storeFreeze } from 'ngrx-store-freeze';
 
 import { environment } from '@env/environment';
@@ -11,9 +12,11 @@ import { initStateFromLocalStorage } from './meta-reducers/init-state-from-local
 import { debug } from './meta-reducers/debug.reducer';
 import { AuthState } from './auth/auth.models';
 import { authReducer } from './auth/auth.reducer';
+import { RouterStateUrl } from './router/router.state';
 
 export const reducers: ActionReducerMap<AppState> = {
-  auth: authReducer
+  auth: authReducer,
+  router: routerReducer
 };
 
 export const metaReducers: MetaReducer<AppState>[] = [
@@ -30,6 +33,12 @@ export const selectAuthState = createFeatureSelector<AppState, AuthState>(
   'auth'
 );
 
+export const selectRouterState = createFeatureSelector<
+  AppState,
+  RouterReducerState<RouterStateUrl>
+>('router');
+
 export interface AppState {
   auth: AuthState;
+  router: RouterReducerState<RouterStateUrl>;
 }
--- a/src/app/examples/crud/books.actions.ts
+++ b/src/app/examples/crud/books.actions.ts
@@ -1,23 +1,14 @@
 import { Action } from '@ngrx/store';
-import { Update } from '@ngrx/entity';
-
 import { Book } from './books.model';
 
 export enum BookActionTypes {
-  ADD_ONE = '[Books] Add One',
-  UPDATE_ONE = '[Books] Update One',
-  DELETE_ONE = '[Books] Delete One',
-  SELECT_ONE = '[Books] Select One'
+  UPSERT_ONE = '[Books] Upsert One',
+  DELETE_ONE = '[Books] Delete One'
 }
 
-export class ActionBooksAddOne implements Action {
-  readonly type = BookActionTypes.ADD_ONE;
+export class ActionBooksUpsertOne implements Action {
+  readonly type = BookActionTypes.UPSERT_ONE;
   constructor(readonly payload: { book: Book }) {}
-}
-
-export class ActionBooksUpdateOne implements Action {
-  readonly type = BookActionTypes.UPDATE_ONE;
-  constructor(readonly payload: { update: Update<Book> }) {}
 }
 
 export class ActionBooksDeleteOne implements Action {
@@ -25,13 +16,4 @@ export class ActionBooksDeleteOne implements Action {
   constructor(readonly payload: { id: string }) {}
 }
 
-export class ActionBooksSelect implements Action {
-  readonly type = BookActionTypes.SELECT_ONE;
-  constructor(readonly payload: { id: string }) {}
-}
-
-export type BookActions =
-  | ActionBooksAddOne
-  | ActionBooksUpdateOne
-  | ActionBooksDeleteOne
-  | ActionBooksSelect;
+export type BookActions = ActionBooksUpsertOne | ActionBooksDeleteOne;
--- a/src/app/examples/crud/books.model.ts
+++ b/src/app/examples/crud/books.model.ts
@@ -7,6 +7,4 @@ export interface Book {
   description: string;
 }
 
-export interface BookState extends EntityState<Book> {
-  selectedBookId: string | null;
-}
+export interface BookState extends EntityState<Book> {}
--- a/src/app/examples/crud/books.selectors.ts
+++ b/src/app/examples/crud/books.selectors.ts
@@ -1,27 +1,23 @@
 import { createSelector } from '@ngrx/store';
 
+import { selectRouterState } from '@app/core';
+
 import { selectExamples, ExamplesState } from '../../examples/examples.state';
 
-import { BookState } from './books.model';
 import { bookAdapter } from './books.reducer';
 
 const { selectEntities, selectAll } = bookAdapter.getSelectors();
-const getSelectedBookId = (state: BookState) => state.selectedBookId;
 
 export const selectBooks = createSelector(
   selectExamples,
   (state: ExamplesState) => state.books
 );
 
-export const selectSelectedBookId = createSelector(
-  selectBooks,
-  getSelectedBookId
-);
-
-export const selectBooksEntities = createSelector(selectBooks, selectEntities);
 export const selectAllBooks = createSelector(selectBooks, selectAll);
+export const selectBooksEntities = createSelector(selectBooks, selectEntities);
+
 export const selectSelectedBook = createSelector(
   selectBooksEntities,
-  selectSelectedBookId,
-  (bookEntities, selectedBookId) => bookEntities[selectedBookId]
+  selectRouterState,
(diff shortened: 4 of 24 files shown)


src/app/examples/examples.state.ts
```diff
@@ -10,25 +15,54 @@ import { bookReducer } from './crud/books.reducer';
...
   books: bookReducer,
-  form: formReducer
+  form: formReducer,
+  router: routerReducer
 };
 
 export const selectExamples = createFeatureSelector<State, ExamplesState>(
   FEATURE_NAME
 );
+export const selectRouterState = createFeatureSelector<
```
timdeschryver: This is the reason why the router is undefined, because we're using [`createFeatureSelector`](https://github.com/ngrx/platform/blob/master/docs/store/selectors.md#createfeatureselector).

> The createFeatureSelector is a convenience method for returning a top level feature state. It returns a typed selector function for a feature slice of state.

And in our case the router is nested inside the examples state, to fix this we'll have to use a "simple" createSelector:
```
export const selectRouterState = createSelector(
  selectExamples,
  (state: ExamplesState) => state.router
);
```

But I think it would be better to move the router state to the top level, then it would be possible to use `createFeatureSelector`.

The reason why I would move this to the top level is because this isn't examples bound and is in my eyes the same level as `auth` and `settings`.

Thoughts?

Jeykairis: Thanks for the explanation and the workaround to get the router state with a selector.

I also thought it makes more sens to get the router state to the top level. So I moved it up.

Now it is nearly working but it should be improved as I need to sync the url with the selectedBookId.

timdeschryver: I think having a `selectedBookId` isn't needed when we're selecting the book based on the url id.

timdeschryver:
@Jeykairis can you you give a heads up when this is ready for a review? :)

src/app/core/core.state.ts
```diff
@@ -30,6 +39,33 @@ export const selectAuthState = createFeatureSelector<AppState, AuthState>(
   'auth'
 );
 
+export const selectRouterState = createFeatureSelector<
```
timdeschryver: Let's make this typed to stay consistent with the rest of the codebase

```
export const selectRouterState = createFeatureSelector<AppState, RouterReducerState>('router');
```

src/app/core/core.state.ts
```diff
@@ -30,6 +39,33 @@ export const selectAuthState = createFeatureSelector<AppState, AuthState>(
...
+  RouterReducerState<RouterStateUrl>
+>('router');
+
+@Injectable()
+export class CustomSerializer implements RouterStateSerializer<RouterStateUrl> {
```
timdeschryver: I think it will be cleaner if we move this to a separate file, thoughts?

src/app/examples/crud/books.actions.ts
```diff
@@ -1,37 +1,19 @@
...
-  UPDATE_ONE = '[Books] Update One',
-  DELETE_ONE = '[Books] Delete One',
-  SELECT_ONE = '[Books] Select One'
+  UPSERT_ONE = '[Books] Upsert One',
+  DELETE_ONE = '[Books] Delete One'
 }
 
-export class ActionBooksAddOne implements Action {
-  readonly type = BookActionTypes.ADD_ONE;
+export class ActionBooksUpsertOne implements Action {
```
timdeschryver: I'm fine with this change, but try to split it up in two PRs the next time 😄

src/app/examples/crud/books.model.ts
```diff
@@ -1,12 +1,6 @@
-import { EntityState } from '@ngrx/entity';
-
 export interface Book {
   id: string;
   title: string;
   author: string;
   description: string;
 }
-
-export interface BookState extends EntityState<Book> {
```
timdeschryver: Having the BookState is cleaner instead of using `EntityState<Book>`.
Also while doing this it's easier if we want to add an extra property later.

```
export interface BookState extends EntityState<Book> {}
```

src/app/examples/crud/books.selectors.ts
```diff
@@ -1,27 +1,27 @@
...
-export const selectSelectedBookId = createSelector(
-  selectBooks,
-  getSelectedBookId
-);
-
-export const selectBooksEntities = createSelector(selectBooks, selectEntities);
 export const selectAllBooks = createSelector(selectBooks, selectAll);
+export const selectBooksEntities = createSelector(selectBooks, selectEntities);
+
+export const selectRouteParameters = createSelector(
```
timdeschryver: This selector isn't being used, thus can be removed.

src/app/examples/examples.module.ts
```diff
@@ -24,12 +24,19 @@ import { BooksEffects } from './crud/books.effects';
...
+  StoreRouterConnectingModule,
+  RouterStateSerializer
+} from '@ngrx/router-store';
 
 @NgModule({
   imports: [
     SharedModule,
     ExamplesRoutingModule,
     StoreModule.forFeature(FEATURE_NAME, reducers),
+    StoreRouterConnectingModule.forRoot({
```
timdeschryver: Move this to `core.module`, the default `stateKey` value is router so we can change this to:

```
StoreRouterConnectingModule.forRoot()
```

Or is it more clear if we would do it like yo had?

src/app/examples/examples.module.ts
```diff
@@ -55,7 +62,10 @@ import { AuthenticatedComponent } from './authenticated/authenticated.component'
     CrudComponent,
     FormComponent
   ],
-  providers: [StockMarketService]
+  providers: [
+    StockMarketService,
+    { provide: RouterStateSerializer, useClass: CustomSerializer }
```
timdeschryver: Same here, move this to `core.module`

src/app/core/core.state.ts
Jeykairis: Yes indeed but I hope it will be remove soon as it should be the default serializer.

src/app/examples/crud/books.actions.ts
Jeykairis: Okay, this change comes to help me workaround another issue 😅

src/app/core/core.state.ts
Jeykairis: Using this won't let provide the `CustomSerializer` in the selector.
So `state` in `(entities, params) => params && entities[params.state.params.id]` only contains `root` and `url`.

src/app/examples/examples.module.ts
Jeykairis: It was indeed a first try to learn that can be cleaned using de default value.

src/app/core/core.state.ts
timdeschryver: Woops you're right, I meant: 

```
export const selectRouterState = createFeatureSelector<AppState, RouterReducerState<RouterStateUrl>>('router');
```

src/app/core/core.state.ts
timdeschryver: `CustomSerializer` isn't the default serializer fron NgRx - https://github.com/ngrx/platform/blob/master/modules/router-store/src/serializer.ts#L23.


It isn't needed anymore to provide a custom serializer because the angular router snapshot became serializable, which was not the case before.

If we would remove the `CustomSerializer` we would have to do `route.firstChild.firstChild.firstChild.params.id` in our selector which isn't really pretty. So I would suggest to keep the `CustomSerializer`.

src/app/core/core.state.ts
timdeschryver: Try changing the string parameter to `roter` or change the `router` property of the `Appstate` to `roter` and you'll notice it 😄 

It's makes sure the string you give with `createFeatureSelector` exists in the state tree.

Jeykairis: Okay, great! Thank you 😄

src/app/examples/crud/books.model.ts
Jeykairis: [This one](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/355/commits/7413986b0b62975f9ac6ff1dfd0ec3abd3ca2711) should be fixed.

tomastrajan:
@Jeykairis looks great, thanks! It's nice that it is now possible to share link to particular book.

The only thing left would be to `rebase` 10 commits into one which follow standard format specified in [Contributing guide](https://github.com/tomastrajan/angular-ngrx-material-starter/blob/master/CONTRIBUTING.md) and we're good to merge!

Jeykairis:
Usually, we use squash and merge by renaming the commit as wanted to keep the master history clean.
On my side I use `git reset --mixed` then commit all changes with the excepted message and then `push -f` to erase the history of the branch.
Is it what you except or do you have more instructions for the `rebase`? 
I also try to `git rebase -i  e96e406f430bb5ccd88879d6e413fea4a0e35d61` and replace all `pick`by `squash` but not sure it's what you want neither.

tomastrajan:
@Jeykairis I could try using githubs `squash and merge`. In my professional project I was usually doing `git rebase -i HEAD^<number-of-commits>` and then replaced `pick` with `fixup`

> `f`, `fixup` = like `squash`, but discard this commit's log message

and rewrite one commit that was `pick` to have the final message. Then it is a single commit PR so it can be just merged. I can try using GitHub's `squash and merge` if you want.

timdeschryver:
I usually use githubs squash and merge for this.
Or `git rebase master -i`, and after this you'll have to force push.

tomastrajan:
@Jeykairis hmm that's strange error. Besides that it is on your branch so that means once you rebase it you force push it to your own origin ( which is your fork ) which should update the PR automatically. Anyway no worries I can just squash and merge.

tomastrajan:
@Jeykairis ah sorry my bad, the `^` means like one so `^^` is like 2 commits but in your case the correct way to do this is `rebase -i HEAD~10`, then `fixup` all besides one, then change commit message of the one left and the `--force` push. Sorry for the confusion :)

Jeykairis:
Thank you @tomastrajan and @timdeschryver for your explanations and letting me the time to explore the rebase tool. 
The `git rebase master -i` first looks to ask me less work by removing commits resolving conflicts with master but next I have to make some fix after the rebase.
Finally, here is the clean commit 😄

tomastrajan:
@Jeykairis amazing, thank you for your effort!

PR #356, opened 2018-09-23
[WIP] add cypress

timdeschryver:
## What:

Introduces cypress to run our e2e tests.
- cypress config and npm scripts are added
- protractor tests are rewritten
- `data-testid` attributes are added in the html to use in the tests, prettier also formatted these files

## Need to clear out
- [x] Is there a way to run the version build with `npm run build:prod`?
- [x] If this is OK what are we going to do with protractor?
- [x] Add to CI?

Issue number: N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,8 @@
     "cy:open": "cypress open",
     "cy:run": "cypress run",
     "precy:ci": "npm run build",
-    "cy:ci": "npm-run-all --parallel --race start:ci \"cy:run -- --config=baseUrl=http://localhost:4000\"",
-    "start:ci": "angular-http-server --path ./dist -p 4000 --silent"
+    "cy:ci": "npm-run-all --parallel --race cy:server \"cy:run -- --config=baseUrl=http://localhost:4000\"",
+    "cy:server": "angular-http-server --path ./dist -p 4000 --silent"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -21,7 +21,12 @@
     "analyze": "npm run clean && npm run build:prod -- --stats-json && webpack-bundle-analyzer ./dist/stats.json",
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate && node .all-contributors-html.js",
-    "contributors:check": "all-contributors check"
+    "contributors:check": "all-contributors check",
+    "cy:open": "cypress open",
+    "cy:run": "cypress run",
+    "precy:ci": "npm run build",
+    "cy:ci": "npm-run-all --parallel --race start:ci \"cy:run -- --config=baseUrl=http://localhost:4000\"",
+    "start:ci": "angular-http-server --path ./dist -p 4000 --silent"
   },
   "husky": {
     "hooks": {
@@ -68,10 +73,14 @@
     "@angular/cli": "~6.1.3",
     "@angular/compiler-cli": "^6.1.2",
     "@angular/language-service": "^6.1.2",
+    "@bahmutov/add-typescript-to-cypress": "^2.0.0",
     "@types/jasmine": "~2.8.3",
     "@types/node": "~9.4.0",
     "all-contributors-cli": "^5.2.1",
+    "angular-http-server": "^1.5.0",
     "codelyzer": "~4.4.2",
+    "cypress": "^3.1.0",
+    "http-server": "^0.11.1",
     "husky": "^1.0.0-rc.13",
     "jasmine-core": "~2.99.1",
     "jasmine-marbles": "^0.3.1",
@@ -84,6 +93,7 @@
     "karma-spec-reporter": "~0.0.31",
     "lint-staged": "^7.2.2",
     "ngrx-store-freeze": "^0.2.4",
+    "npm-run-all": "^4.1.3",
     "prettier": "^1.7.4",
     "protractor": "^5.3.2",
     "rimraf": "^2.6.2",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "clean": "rimraf ./dist/",
     "server": "cd dist && http-server",
     "prod": "npm run clean && npm run build:prod && npm run server",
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -21,7 +21,12 @@
     "analyze": "npm run clean && npm run build:prod -- --stats-json && webpack-bundle-analyzer ./dist/stats.json",
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate && node .all-contributors-html.js",
-    "contributors:check": "all-contributors check"
+    "contributors:check": "all-contributors check",
+    "cy:open": "cypress open",
+    "cy:run": "cypress run",
+    "precy:ci": "npm run build",
+    "cy:ci": "npm-run-all --parallel --race start:ci \"cy:run -- --config=baseUrl=http://localhost:4000\"",
+    "start:ci": "angular-http-server --path ./dist -p 4000 --silent"
```
tomastrajan: Could we prefix everything with `cy:` so that it is obvious that they belong together?

tomastrajan:
@timdeschryver looks very cool! Can we also run it as a part of `npm run ci` ?

package.json
timdeschryver: Yes!

timdeschryver:
> @timdeschryver looks very cool! Can we also run it as a part of npm run ci ?

We definitely can, but wanted to hear your thoughts on this, as well as the bullet points under "Need to clear out".

package.json
```diff
@@ -21,7 +21,12 @@
     "analyze": "npm run clean && npm run build:prod -- --stats-json && webpack-bundle-analyzer ./dist/stats.json",
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate && node .all-contributors-html.js",
-    "contributors:check": "all-contributors check"
+    "contributors:check": "all-contributors check",
+    "cy:open": "cypress open",
+    "cy:run": "cypress run",
+    "precy:ci": "npm run build",
+    "cy:ci": "npm-run-all --parallel --race cy:server \"cy:run -- --config=baseUrl=http://localhost:4000\"",
```
tomastrajan: Do I understand correctly that this tries to check if there is already existing prod build in `dist` so that it can skip rebuilding but will rebuild with `cy:run` if `./dist` is empty ?

timdeschryver: No this will always build the app, once it's build it will start a server on port 4000 which is used by cypress to test the app.

timdeschryver:
> Is there a way to run the version build with npm run build:prod?

To build the app I tried using `npm run build:prod` but then the translations wouldn't come through because it searches the translations with

```
return new TranslateHttpLoader(
    http,
    `${environment.i18nPrefix}/assets/i18n/`,
    '.json'
  );
```

Thus it can't find the translations.

If I build the app with `--deploy-url /angular-ngrx-material-starter/ --base-href /angular-ngrx-material-starter` flags, I can't seem to run the app which is normal. The `angular-ngrx-material-starter` is needed for the github pages.

package.json
tomastrajan: Hmm, I am just trying to wrap my head around that the `npm-run-all` with

> `--race`  - Set the flag to kill all tasks when a task **finished with zero**. This option is valid only with 'parallel' option.

and

> Well-behaved UNIX commands, programs, and utilities **return a 0** exit code upon successful completion, though there are some exceptions.

So then only one of them (the faster will complete) ? Why do we try both ?

tomastrajan:
@timdeschryver could cypress be configured so it also run with the context path of `/angular-ngrx-material-starter/` ? 

Also...
> I can't seem to run the app which is normal. 

Can you please elaborate a bit more why this is the case?
I have no experience with cypress so maybe I am missing obvious things...

timdeschryver:
This has nothing to do with Cypress but with "hosting" the application yourself. 
For example with `http-server`  or `angular-http-server`.

I'll take another look later to see if it's possible. In the past I wasn't succesful in finding a solution 😅

package.json
timdeschryver: If we wouldn't do this and simply do "npm run cy:server && npm run cy:run", the `cy:run` command will never be run because the server command will not return an exit code (it just sits there and waits).

With `npm-run-all` you can run multiple jobs in parallel, so in our case both the `server` and `run` command will execute. Like I said before the `server` command just sits there, but the `cy:run` command will return an exit code when all the tests have run. When it retrieves an exit code it will also shut down the server.

tomastrajan: Ah ok, I understand now, it's a way to tear down server when the tests are finished. Clever! 

I guess I got confused because I was not aware of expected lifetimes of those commands, now it all make sense. Maybe we can also add a small `Cypress testing` section in the `README.md` with the happy path usage scenario and gotchas ;)

tomastrajan:
This could work?
```
// cypress-server.js
const express = require('express');
const app = express();
app.use('/angular-ngrx-material-starter', express.static(__dirname + '/dist'));
app.listen(4200);
```

timdeschryver:
I'll try it later today and keep you posted.

timdeschryver:
It now tries to find the `*.js`, and `*.css` files, but is looking at the wrong path.

timdeschryver:
It should work with the last commit but I don't like it that we have to do the following in the cypress tests. Also I think this might crash the `cy:open` command.

```
cy.visit('/#about');
```

Another solution would be to not use the `prod` flag, or to change the localization paths if we're running the cypress ci tests. I think to do the latter we would have to introduce a new environment?

timdeschryver:
`/#/about` works.

With giving it a second look with a fresh head I think we don't need a another env file. I didn't notice we were also hashing the router in dev mode. Our cypress tests runs fine in dev mode and prod mode, so I don't see a need to "complicate" things and add another config.

So I think all the things we needed to clear out are done:
- we are building our app in `prod` mode and using an express server to run the app
- we've added it to the ci
- we're going to use protractor and cypress side by side to give a good comparison 

If you're OK with this setup I can add it to the README.

tomastrajan:
@timdeschryver yes, I think it is great like this, please update the README.md and we can merge this :)

singhcharanjit:
I added a new module, added new folder under i18n folder and also added TranslateModule.forChild() in imports of new module. Somehow it is not loading translation for new module.

What am I missing?

PR #357, opened 2018-09-24
feat(fontawesome): upgrade to the new angular-fontawesome

Jeykairis:
## What:

The package [@fortawesome/fontawesome-free-webfonts](https://www.npmjs.com/package/@fortawesome/fontawesome-free-webfonts) is deprecated and does not contain all new icons.

As they provide a [new package based on inline svg specially for angular](https://www.npmjs.com/package/@fortawesome/angular-fontawesome) it looks like the best way to use font awesome.

So here is the proposition with the upgrade of first icons.

Issue number: N/A


tomastrajan:
Generally agreed, just one problem,
this change bloats `vendor` bundle by approximately `1mb`, can this be optimized?

src/app/app.component.html
```diff
@@ -48,17 +49,17 @@
         <button *ngIf="isAuthenticated"
                 mat-icon-button
                 [matMenuTriggerFor]="toolbarUserMenu">
-          <mat-icon fontSet="fas" fontIcon="fa-user-circle"></mat-icon>
+          <fa-icon icon="user-circle"></fa-icon>
         </button>
         <mat-menu #toolbarUserMenu="matMenu">
           <button mat-menu-item (click)="onLogoutClick()">
-            <mat-icon fontSet="fas" fontIcon="fa-power-off"></mat-icon>
+            <mat-icon><fa-icon icon="power-off"></fa-icon></mat-icon>
```
tomastrajan: Can we remove this `mat-icon`, and search for all occurrences and remove them? Can we also remove `MatIconModule` from the `SharedModule` ?

Jeykairis: This specific `mat-icon` preserve the design of the `mat-menu-item`. So we should keep `MatIconModule` or copy the style applied for mat-menu-item which seems not ideal.

This is a bit annoying and the size of vendor increase with each icon added. First I hope the tree shaking remove unused one but we have to specify each one we want to use.

Regarding both previous points, do you believe inline svg icons with this library is better than using the font option with scss import?

tomastrajan: I see, yeah ok let's move with the `svg` icons as in current version of the PR, please increase the sizes of the `vendor` budgets in `angular.json` appropriately ( rounded to 50kb ), fix merge conflicts and we can merge this. 

I think it is the better approach in the long term where we can specify which icons we're interested in and also as you said, the previous usage style was deprecated so we should move on...

tomastrajan:
@Jeykairis thank you for changes, well done !

tomastrajan:
@Jeykairis will do a new release beginning of next week ;)

PR #360, opened 2018-09-29
feat(): add UI notification for client and HTTP errors

hhubik:
@timdeschryver any thoughts on this?  I tried to keep it simple: just extended the default error handler to show a snackbar and then hand the error off to the default error handler for logging.  Also added a simple test of localStorage availability.


timdeschryver:
Great, thanks in advance! 
I'll take a look during the week.

src/app/app.component.ts
```diff
@@ -169,4 +171,18 @@ export class AppComponent implements OnInit, OnDestroy {
...
     });
   }
+
+  /** Verifies that localStorage can be written to and read from */
+  private testLocalStorage() {
+    const testValue = 'testValue';
+    const testKey = 'testKey';
+    let retrievedValue: string;
+
+    localStorage.setItem(testKey, testValue);
```
timdeschryver: I think adding wrapping this in a try-catch would be safer?
Also, should we also test if `localStorage` exists?

src/app/app.component.ts
```diff
@@ -169,4 +171,18 @@ export class AppComponent implements OnInit, OnDestroy {
...
+    retrievedValue = localStorage.getItem(testKey);
+
+    if (!retrievedValue || retrievedValue !== testValue) {
```
timdeschryver: We can leave remove the `!retrievedValue` check here, because it's redundant.

timdeschryver:
This is how it looks like for the ones wondering:

image

src/app/core/http-interceptors/http-error.interceptor.ts
```diff
@@ -0,0 +1,29 @@
...
+@Injectable()
+export class HttpErrorInterceptor implements HttpInterceptor {
+  constructor(private injector: Injector) {}
+
+  intercept(
+    request: HttpRequest<any>,
+    next: HttpHandler
+  ): Observable<HttpEvent<any>> {
+    return next.handle(request).pipe(
+      tap(null, httpErrorResponse => {
```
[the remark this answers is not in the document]
tomastrajan: @timdeschryver the next step  would be `NotificationService` in core which will display styled toasts with `info`, `warn`, `error` API ;)

src/app/app.component.ts
hhubik: I will add the test of existence of `localStorage`.  As to wrapping this in `try-catch`, that seems to defeat the purpose of a global error handler.  What would we do in the `catch` besides re-throwing the exception?

tomastrajan: Maybe this could be in the `core/local-storage.service` ?

hhubik: @tomastrajan will do!

src/app/core/http-interceptors/http-error.interceptor.ts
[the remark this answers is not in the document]
timdeschryver: I think we can leave it as it is, with the comment @tomastrajan made.

What I meant was, doing something like this: 

```ts
 return next.handle(request).pipe(
      catchError(err => {
       ... handle error here...
      })
    );
```

hhubik:
Sorry, I could not figure out how to sync my fork and my local repo with the starter project.  I am closing this and will open a new PR against a new fork.

PR #367, opened 2018-10-01
test(auth): add unit test for auth selectors, closes #336

### LINKED ISSUE #336: Create unit tests for auth selectors
Create tests for [auth.seelctors.ts](https://github.com/tomastrajan/angular-ngrx-material-starter/blob/master/src/app/core/auth/auth.selectors.ts) in `auth.selectors.spec.ts` file by **following already existing tests**, the official Angular testing guide or feel free to suggest improvements!

Don't forget to reference this issue in the commit message (eg `test(auth): add unit tests for <something>, closes #<issue-number>`)

erhise:
## What:

There is really no difference between selectAuth and selectAuthState, both are memoized selectors and they are returning the same values.

I did just add another selector for good practice, even though the AuthState is very small, so today there is no practical difference. Therefore I did not take it upon myself to change code where selectAuth is used.

But if the scenario instead would be
`auth: {
  isAuthenticated: false,
  someOtherProp: false
}`
and you only are interested in isAuthenticated. The selectAuth would result in multiple subscription triggers if someOtherProp would change. 

Issue number: #336

### DIFF
--- a/src/app/core/auth/auth.selectors.spec.ts
+++ b/src/app/core/auth/auth.selectors.spec.ts
@@ -1,0 +1,19 @@
+import { selectAuth, selectIsAuthenticated } from './auth.selectors';
+
+const createAuthState = ({ isAuthenticated = false } = {}) => ({
+  auth: {
+    isAuthenticated
+  }
+});
+
+describe('Auth Selectors', () => {
+  const state = createAuthState();
+
+  it('selectAuth', () => {
+    expect(selectAuth(state)).toBe(state.auth);
+  });
+
+  it('selectIsAuthenticated', () => {
+    expect(selectIsAuthenticated(state)).toBe(state.auth.isAuthenticated);
+  });
+});
--- a/src/app/core/auth/auth.selectors.spec.ts
+++ b/src/app/core/auth/auth.selectors.spec.ts
@@ -1,19 +1,21 @@
 import { selectAuth, selectIsAuthenticated } from './auth.selectors';
 
-const createAuthState = ({ isAuthenticated = false } = {}) => ({
-  auth: {
-    isAuthenticated
-  }
-});
-
 describe('Auth Selectors', () => {
-  const state = createAuthState();
-
   it('selectAuth', () => {
+    const state = createAuthState();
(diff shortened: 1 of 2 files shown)


src/app/core/auth/auth.selectors.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { selectAuth, selectIsAuthenticated } from './auth.selectors';
+
+const createAuthState = ({ isAuthenticated = false } = {}) => ({
```
timdeschryver: Move this to the bottom of the page to stay consistent.

src/app/core/auth/auth.selectors.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  auth: {
+    isAuthenticated
+  }
+});
+
+describe('Auth Selectors', () => {
+  const state = createAuthState();
```
timdeschryver: Recreate this state in every test, instead of once globally.

src/app/core/auth/auth.selectors.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  it('selectAuth', () => {
+    expect(selectAuth(state)).toBe(state.auth);
+  });
+
+  it('selectIsAuthenticated', () => {
+    expect(selectIsAuthenticated(state)).toBe(state.auth.isAuthenticated);
```
timdeschryver: Wouldn't it be better to use `.toBe(false)` here?

src/app/core/auth/auth.selectors.spec.ts
erhise: This itself gives lint error "variable 'createAuthState' used before declaration", unless if I didn't misunderstood you.

timdeschryver: I think if you encounter this lint error, making it a function would resolve it.

timdeschryver:
PS: feel free to add yourself to the list of contributors `npm run contributors:add -- erhise test` and `npm run contributors:generate`.

https://www.npmjs.com/package/all-contributors-cli

PR #375, opened 2018-10-01
feat: angular flex layout (WIP)

rpang77:
## What:

Issue number: N/A


tomastrajan:
Hi @rpang77 !

Thank you for your effort! Could you please provide some more details about your motivation for switching to `flex-layout`? Also it would be very nice to provide quick summary of how bootstrap responsive features map to that of the flex.

Haven't checked it in quite some time but the way I remember it is, that id doesn't really support responsive grid out of the box and you have to define it yourself. Eg 12 column responsive grid.

tomastrajan:
We will not be moving forward with this at the moment. @rpang77 thank you for your efforts and until next time!

PR #378, opened 2018-10-02
fix(form): restore the form state consistently with the local storage

Jeykairis:
## What:

Restore the form state consistently with the local storage.

Issue number: N/A


tomastrajan:
@Jeykairis @timdeschryver thanks a lot!

PR #383, opened 2018-10-05
fix on app-routing

enderkey:
## What:

Issue number: N/A


tomastrajan:
Hi @enderkey , can you please provide bit more description? Also, it seems to be breaking CI build?

PR #384, opened 2018-10-05
refactor(features): separate data from the template

Jeykairis:
## What:

Separate data from the template.

Issue number: N/A


src/app/static/features/features.component.html
```diff
@@ -5,225 +5,26 @@ <h1 class="main-heading">{{ 'anms.features.title' | translate }}</h1>
...
-            {{ 'anms.features.documentation' | translate }}
-          </a>
-        </mat-card-actions>
-      </mat-card>
-    </div>
-    <div class="col-md-6 col-lg-4" [ngClass]="routeAnimationsElements">
-      <mat-card>
-        <mat-card-title><code>{{versions.ngxtranslate}}</code> I18n
+        <mat-card-title>
+          <ng-container *ngIf="feature.version; else translateName;">
```
timdeschryver: What about just doing the following?

```html
<mat-card-title>
          <code *ngIf="feature.version">{{ feature.version }}</code>{{ feature.name | translate }}
</mat-card-title>
```

Jeykairis: It's way simpler 😄

src/app/static/features/features.component.ts
```diff
@@ -1,16 +1,17 @@
 import { Component, OnInit } from '@angular/core';
 
-import { environment as env } from '@env/environment';
 import { ROUTE_ANIMATIONS_ELEMENTS } from '@app/core';
 
+import features from './features';
```
tomastrajan: Maybe `import { features } from './features-data'` or features-list or something?

and then in comp `features = features` or `features = featuresData`. Just because the `features = features.features;` is a bit goofy 😹

Also maybe we can provide `export interface Feature {}` with mandatory and optional fields so that it is easy to add more for the feature contributors ;)

Jeykairis: Totally agreed, this one should be cleaner. I just found the way to export data as typed const instead of default.

tomastrajan:
@Jeykairis thank you!

PR #390, opened 2018-10-14
test(form): add form effects tests

zetsnotdead:
Add missing form effects tests

### DIFF
--- a/src/app/examples/form/form.effects.spec.ts
+++ b/src/app/examples/form/form.effects.spec.ts
@@ -1,0 +1,50 @@
+import { LocalStorageService } from '@app/core';
+import { Actions, getEffectsMetadata } from '@ngrx/effects';
+import { cold } from 'jasmine-marbles';
+import { EMPTY } from 'rxjs';
+import { ActionFormUpdate } from './form.actions';
+import { FormEffects, FORM_KEY } from './form.effects';
+import { Form } from './form.model';
+
+describe('FormEffects', () => {
+  let localStorageService: LocalStorageService;
+
+  beforeEach(() => {
+    localStorageService = jasmine.createSpyObj('LocalStorageService', [
+      'setItem'
+    ]);
+  });
+
+  describe('persistForm', () => {
+    it('should not dispatch any action', () => {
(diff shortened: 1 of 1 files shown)


tomastrajan:
@zetsnotdead is there a reason why this was implemented with `TestBed` compared to your [other implementation](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/389/files) ?

zetsnotdead:
I implemented it before reading your comment on the other PR. Going with the TestBed was more inline with other specs in the repo, and I considered going for TestBedless specs a bit experimental :laughing:

src/app/examples/form/form.effects.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    formEffect = TestBed.get(FormEffects);
+  });
+
+  it('should be created', () => {
+    expect(formEffect).toBeTruthy();
+  });
+
+  it('should not dispatch any action', () => {
+    metadata = getEffectsMetadata(formEffect);
+    expect(metadata.persistForm).toEqual({ dispatch: false });
```
timdeschryver: Since we're talking about `TestBed`, using `beforeEach` instead of functions, ...

I personally would want to remove all of the `getEffectsMetadata` tests.
I don't think they add much value and are made with a "I have to write tests, so I'll write tests" mindset.

Thoughts?

tomastrajan: @timdeschryver hmm, I can see your point. On the other hand, we are testing for the cases when `@Effect` is supposed to dispatch some action so I guess it's also makes sense to "enforce" that it doesn't. Without such a test change in implementation can cause weird business errors when `@Effect` is still dispatching some action which is not supposed to which leads to unexpected state changes?

timdeschryver: Won't that be tested with `expect(foo).toBeObservable(bar)`?

zetsnotdead:
@tomastrajan @timdeschryver should I remove the TestBed here as well?

zetsnotdead:
Testbed is gone :)

PR #393, opened 2018-10-15
feat(): add UI notification for client and HTTP errors

### LINKED ISSUE #321: Handle Http Service errors in StockMarket example
Per the discussion in #318 ...

> Http Errors (e.g. server unavailable, timeouts, etc. These could be handled for example using the HttpInterceptor as explained in this [video](https://youtu.be/tS1CG4Qx5ow).

hhubik:
Added application-wide error handler that shows a UI notification
for any client or HTTP error before passing the error to
the default Angular ErrorHandler.

closes #321

Please see #360 for review comments for the previous version of my changes.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -14,7 +14,8 @@ import {
   TitleService,
   selectAuth,
   routeAnimations,
-  AppState
+  AppState,
+  LocalStorageService
 } from '@app/core';
 import { environment as env } from '@env/environment';
 
@@ -64,7 +65,8 @@ export class AppComponent implements OnInit, OnDestroy {
     private router: Router,
     private titleService: TitleService,
     private animationService: AnimationsService,
-    private translate: TranslateService
+    private translate: TranslateService,
+    private storageService: LocalStorageService
   ) {}
 
   private static trackPageView(event: NavigationEnd) {
@@ -81,6 +83,7 @@ export class AppComponent implements OnInit, OnDestroy {
     this.subscribeToSettings();
     this.subscribeToIsAuthenticated();
     this.subscribeToRouterEvents();
+    this.storageService.testLocalStorage();
   }
 
   ngOnDestroy(): void {
--- a/src/app/core/local-storage/local-storage.service.ts
+++ b/src/app/core/local-storage/local-storage.service.ts
@@ -45,4 +45,24 @@ export class LocalStorageService {
   getItem(key: string) {
     return JSON.parse(localStorage.getItem(`${APP_PREFIX}${key}`));
   }
+
+  removeItem(key: string) {
+    localStorage.removeItem(`${APP_PREFIX}${key}`);
+  }
+
+  /** Tests that localStorage exists, can be written to, and read from. */
+  testLocalStorage() {
+    const testValue = 'testValue';
+    const testKey = 'testKey';
+    let retrievedValue: string;
+    const errorMessage = 'localStorage did not return expected value';
+
+    this.setItem(testKey, testValue);
+    retrievedValue = this.getItem(testKey);
+    this.removeItem(testKey);
+
+    if (retrievedValue !== testValue) {
+      throw new Error(errorMessage);
+    }
+  }
 }
--- a/src/app/app.component.scss-theme.scss
+++ b/src/app/app.component.scss-theme.scss
@@ -3,6 +3,7 @@
 @mixin anms-app-component-theme($theme) {
   $primary: map-get($theme, primary);
   $accent: map-get($theme, accent);
+  $warn: map-get($theme, warn);
 
   mat-sidenav-container {
     mat-toolbar {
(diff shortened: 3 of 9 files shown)


src/app/core/local-storage/local-storage.service.ts
```diff
@@ -45,4 +45,32 @@ export class LocalStorageService {
   getItem(key: string) {
     return JSON.parse(localStorage.getItem(`${APP_PREFIX}${key}`));
   }
+
+  /** Returns true if localStorage is defined, otherwise false. */
+  localStorageExists(): boolean {
+    if (localStorage) {
```
timdeschryver: Just do `return localStorage`

hhubik: @timdeschryver  - When I try that, I get the following typescript error:
`Type 'Storage' is not assignable to type 'boolean'.`

hhubik: @timdeschryver I can do this though, if you'd like
`return (localStorage) ? true : false;`

timdeschryver: what about `Boolean(localStorage)`?

hhubik: Yes, `Boolean(localStorage)` seems to work.

However, after reading among other things [this SO answer](https://stackoverflow.com/questions/16427636/check-if-localstorage-is-available), I think that checking for localStorage existence is redundant with what the method `testLocalStorage` method does.  The only thing I would change is to remove the test item the method creates.  The method would then look like this:

```typescript
  /** Tests that localStorage exists, can be written to, and read from. */
  testLocalStorage() {
    const testValue = 'testValue';
    const testKey = 'testKey';
    let retrievedValue: string;
    const errorMessage = 'localStorage did not return expected value';

    this.setItem(testKey, testValue);
    retrievedValue = this.getItem(testKey);
    this.removeItem(testKey);

    if (retrievedValue !== testValue) {
      throw new Error(errorMessage);
    }
  }
```
If `localStorage` is unavailable/undefined for any reason, the `set/get/remove` methods will
throw an exception that will be handled by the app error handler (snackbar message displayed and
errors logged in console), which is all we want.

hhubik:
OK. I have just pushed in the changes I had outlined above. Let me know what you think.

src/app/core/local-storage/local-storage.service.ts
hhubik: OK. I have just pushed in the changes I had outlined above. Let me know what you think.

src/app/app.component.ts
```diff
@@ -81,6 +83,7 @@ export class AppComponent implements OnInit, OnDestroy {
     this.subscribeToSettings();
     this.subscribeToIsAuthenticated();
     this.subscribeToRouterEvents();
+    this.storageService.testLocalStorage();
```
tomastrajan: @hhubik ok so if I understand correctly, we use `testLocalStorage();` and if there is no `localStorage` then this will throw exception but it will not kill the application because it will be handled by the `AppErrorHandler` right?

What would be also nice and can be a part of a new issue is to enhance `local-storage.service.ts` to still throw if the localStorage is missing but fall back to in memory object for the functionality for the time application runs (eg `fallbackStorage = {} as any `);

hhubik: @tomastrajan That is correct. Any exception, including one due to unavailability of `localStorage`, will cause an exception that will (1) result in a snackbar popup and (2) be logged in console. The app will not die.  I tested that as follows:
1. add `throw new Error('localStorageException');` line to `testLocalStorage` method.
2. verified that the snack bar pops up.
3. the following is in the console:
```
AppComponent_Host.ngfactory.js? [sm]:1 ERROR Error: localStorageException
    at LocalStorageService.push../src/app/core/local-storage/local-storage.service.ts.LocalStorageService.testLocalStorage (local-storage.service.ts:60)
    at AppComponent.push../src/app/app.component.ts.AppComponent.ngOnInit (app.component.ts:86)
    at checkAndUpdateDirectiveInline (core.js:18287)
    at checkAndUpdateNodeInline (core.js:19551)
    at checkAndUpdateNode (core.js:19513)
    at debugCheckAndUpdateNode (core.js:20147)
    at debugCheckDirectivesFn (core.js:20107)
    at Object.eval [as updateDirectives] (AppComponent_Host.ngfactory.js? [sm]:1)
    at Object.debugUpdateDirectives [as updateDirectives] (core.js:20099)
    at checkAndUpdateView (core.js:19495)

View_AppComponent_Host_0 | @ | AppComponent_Host.ngfactory.js? [sm]:1
-- | -- | --
  | proxyClass | @ | compiler.js:10198
  | push../node_modules/@angular/core/fesm5/core.js.DebugContext_.logError | @ | core.js:20351
  | push../node_modules/@angular/core/fesm5/core.js.ErrorHandler.handleError | @ | core.js:12326
  | push../src/app/core/error-handler/app-error-handler.service.ts.AppErrorHandler.handleError | @ | app-error-handler.service.ts:24
  | (anonymous)
    ...
```
4. After this, I can still navigate to different pages (e.g. Examples) in the app.

tomastrajan: @hhubik thanks for explanation!

timdeschryver:
Sorry I was a bit busy the last couple of days. Just wanted to say kudos to you @hhubik with this 👍

PR #394, opened 2018-10-15
test(todos): remove TestBed from todos effects specs

zetsnotdead:
Refactored todos effects specs as discussed in #389


timdeschryver:
Just on small nitpick, you could also just do `const actions = new Actions();` to create an `EMPTY` actions stream.

PR #395, opened 2018-10-15
test(settings): remove TestBed from settings effects specs

zetsnotdead:
Refactored settings effects specs as discussed in #389

### DIFF
--- a/src/app/settings/settings.effects.spec.ts
+++ b/src/app/settings/settings.effects.spec.ts
@@ -1,52 +1,36 @@
-import { TestBed } from '@angular/core/testing';
-import { provideMockActions } from '@ngrx/effects/testing';
-import { Action } from '@ngrx/store';
-import { EffectsMetadata, getEffectsMetadata } from '@ngrx/effects';
-import { Observable } from 'rxjs';
-import { cold } from 'jasmine-marbles';
-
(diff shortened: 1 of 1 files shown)


src/app/settings/settings.effects.spec.ts
```diff
@@ -73,16 +63,4 @@ describe('SettingsEffects', () => {
       );
     });
   });
-
-  it('should not call methods on AnimationsService and LocalStorageService for other actions', () => {
```
timdeschryver: I think this test is removed by accident? Or was this intended?

tomastrajan: @zetsnotdead pls have a look :)

zetsnotdead: I actually deleted it deliberately. My reason being two fold:
- no other effect spec tests for actions that do nothing, so this test seemed a little out of place
- if we tests for all actions that trigger an event, can't we just assume that all other actions are a noop

If this reasoning does not seem right I will revert it.

timdeschryver: I think I have the same mindset as @zetsnotdead here 😄 
Especially on the 2nd point.

tomastrajan: @zetsnotdead @timdeschryver  thanks for clarification!

PR #401, opened 2018-10-20
fix(app): Double scrollbar on page transition in long pages. fixes #400

### LINKED ISSUE #400: Double scrollbar on page transition in long pages (regression)
## Minimal reproduction of the bug with instructions:
Navigate to page with long content like features or about shows secondary inner scrollbar

## Expected behavior:
Single outer scrollbar

anms-issue

GrzegorzStanczyk:
## What:
I changed overflow-y axis from auto to hidden. And also I changed overflow-x and -y to one overflow property.
I checked if it works correctly in Chrome, IE and Edge.

Issue number:  #400

### DIFF
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -70,8 +70,7 @@ mat-sidenav-container {
     .content {
       flex: 1 0 auto;
       margin-top: 64px;
-      overflow-y: auto;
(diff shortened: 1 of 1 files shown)


tomastrajan:
@GrzegorzStanczyk I tried it out and it seems to be working great, thank you very much!

PR #404, opened 2018-10-23
refactor(notifications): Implemented UI Notifications Service

### LINKED ISSUE #377: Implement Notification Service
Service to provide UI notifications of the following types:
* info
* warn
* error

Per @tomastrajan  suggestion https://github.com/tomastrajan/angular-ngrx-material-starter/pull/360#discussion_r221825600

Should be done after #321 is closed.

hhubik:
Provides a consistent mechanism for displaying self-closing Error, Warning, and Information notifications.

closes #377

### DIFF
--- a/src/app/app.component.scss-theme.scss
+++ b/src/app/app.component.scss-theme.scss
@@ -51,11 +51,20 @@
   .error-notification-overlay {
     color: mat-color($primary, default-contrast);
     background-color: mat-color($warn);
+  }
 
-    button {
-      border-radius: 2px;
-      background-color: mat-color($accent);
-      color: mat-color($accent, default-contrast);
-    }
+  .information-notification-overlay {
+    color: mat-color($primary, default-contrast);
+    background-color: mat-color($primary);
+  }
+
+  .success-notification-overlay {
+    color: mat-color($primary, default-contrast);
+    background-color: $success-color;
+  }
+
+  .warning-notification-overlay {
+    color: mat-color($primary, default-contrast);
+    background-color: $warning-color;
   }
 }
--- a/src/app/core/core.module.ts
+++ b/src/app/core/core.module.ts
@@ -22,6 +22,7 @@ import {
   RouterStateSerializer
 } from '@ngrx/router-store';
 import { CustomSerializer } from './router/custom-serializer';
+import { NotificationService } from './notifications/notification.service';
 
 @NgModule({
   imports: [
@@ -50,6 +51,7 @@ import { CustomSerializer } from './router/custom-serializer';
   ],
   declarations: [],
   providers: [
+    NotificationService,
     LocalStorageService,
     AuthGuardService,
     AnimationsService,
--- a/src/app/core/error-handler/app-error-handler.service.ts
+++ b/src/app/core/error-handler/app-error-handler.service.ts
@@ -1,14 +1,15 @@
-import { Injectable, ErrorHandler, NgZone } from '@angular/core';
+import { Injectable, ErrorHandler } from '@angular/core';
 import { HttpErrorResponse } from '@angular/common/http';
-import { MatSnackBar } from '@angular/material/snack-bar';
 import { environment } from '@env/environment';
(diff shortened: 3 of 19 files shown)


src/app/core/notifications/notifications.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor(
+    private readonly snackBar: MatSnackBar,
+    private readonly zone: NgZone
+  ) {}
+
+  /** Displays a self-closing UI notification with a given message.
+   * @param {string} message - Notification message.
+   * @param {NotificationStyles=} style - Notification style, Information by default.
+   */
+  show(message: string, style?: NotificationStyles) {
```
tomastrajan: Hi @hhubik !

Thanks for your PR! What is your opinion about providing `.info()`, `.warn()`, `.error()`, `.success()`, ... etc convenience methods instead of passing `enum` value ?

hhubik: Hi @tomastrajan I don't have a strong preference one way or another.  I actually briefly considered doing what you suggest, but then decided to do it this way, probably because in my previous life I did C#.  For example, to show a message dialog in C#, you call `MessageBox.Show()` method and control the appearance and content of the message box with input parameters to `Show()`.

If the Javascript/Typescript convention is to use separate methods, I am happy to do that.  Just let me know.  

In general, I prefer to use verbs in methods names.  So I would probably do `.showInfo()` etc.  Is that OK?

tomastrajan: @hhubik thank you for your explanation. The way I think about the notifications is like a more fancy console. That means it kinda follow console API like info, warn, error, but also adds more user specific stuff like `success`. In terms of styling it could be something like this...

info: $primary color from material theme
error: $warn from material ( don't remember the exact name)
success: we have to define another variable for this ( pastel green )
warn: we have to define another variable for this ( pastel yellow )

we should define these new variables globalin `styles-variables.scss` and import wherever needed ( in the `-scss-theme.scss` files.

Please let me know if you have more questions ;)

hhubik: @tomastrajan - I have found a few examples of notification services for Web UI.  They all use your approach.  I will change my code to match that.  Before I do that though, would it make more sense to just use an off-the-shelf solution such as [angular2-notifications](https://www.npmjs.com/package/angular2-notifications)?

tomastrajan: @hhubik I would go with our own lightweight notification service :)

hhubik: @tomastrajan OK.  I am making the changes to the code now.  Could you please elaborate on "we have to define another variable for this ( pastel green )".  Is there an example of how this should be done? Thanks!

tomastrajan: @hhubik so for the success and warning notifications, you will need to define color which are cannot be defined as a part of the theme so you have to define them. 

Themes doesn't support hanging additional properties so for now, we have to define these colors globally in the `styles-variables.scss`. 

They can be called `$success: mat-palette($mat-light-green);` and `$warning: mat-palette($mat-yellow, 400);`....

Then you can use them in the component where you implement notifications (preferably a new component) which will be used in the `app.component.html` template. In terms of applying color, please use additional theme file as for every component (all of them have these files: ts, html, scss and scss-theme.scss)

hhubik:
@tomastrajan  How is this?  Was not completely sure about the SASS/theming stuff.

tomastrajan:
@hhubik i think it looks pretty great!

tomastrajan:
Hey @hhubik ! Check out [this commit](https://github.com/tomastrajan/angular-ngrx-material-starter/commit/32688dfbdd24024832c94e166e6231746df82e24) with various minor improvements on your great PR ! Also you can check how it is possible to handle colors in styles variables and how they can be consumed in theme files ;)

Thanks again!

hhubik:
@tomastrajan I checked it out.  Makes sense, as far as I can tell  ;)  Thanks for the feedback!

PR #411, opened 2018-10-30
test(Form): add happy path

timdeschryver:
## What:
Added a happy path test for the form example.
This is done with [ngx-test-library](https://github.com/timdeschryver/ngx-testing-library).

Issue number: N/A

### DIFF
--- a/src/app/examples/form/components/form.component.spec.ts
+++ b/src/app/examples/form/components/form.component.spec.ts
@@ -13,11 +13,11 @@ import { ActionFormUpdate } from '../form.actions';
 
 fdescribe('FormComponent', () => {
   let store: MockStore<State>;
-  let renderResult: RenderResult;
+  let component: RenderResult;
   let dispatchSpy: jasmine.Spy;
 
   beforeEach(async () => {
-    renderResult = await createComponent<FormComponent>(
+    component = await createComponent<FormComponent>(
       '<anms-form></anms-form>',
       {
         declarations: [FormComponent],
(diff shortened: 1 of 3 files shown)


src/app/examples/form/components/form.component.spec.ts
```diff
@@ -1,35 +1,104 @@
...
+        password: 'mysuperawesomeandsecurepassword',
+        email: 'foo@bar.baz',
+        description: 'no description needed here',
+        requestGift: true,
+        birthday: new Date(1991, 12, 31, 0, 0, 0),
+        rating: 0
+      }
+    });
+    const { birthday, ...formValue } = payload.form;
+    expect({ ...dispatchSpy.calls.mostRecent().args[0] }).toEqual({
```
timdeschryver: If someone got a cleaner way of doing this, feel free to chip in 😅 
Problems I encountered:
- JS Object vs TS Class
- Birthday is a date object, this was giving some troubles; hence the check if it's filled in, instead of comparing both values.

src/app/examples/form/components/form.component.spec.ts
```diff
@@ -1,35 +1,104 @@
...
+  let dispatchSpy: jasmine.Spy;
 
-  beforeEach(async(() => {
-    TestBed.configureTestingModule({
-      imports: [TestingModule],
-      declarations: [FormComponent],
-      providers: [NotificationService]
-    }).compileComponents();
+  beforeEach(async () => {
+    renderResult = await createComponent<FormComponent>(
```
tomastrajan: Maybe call this `renderedComponent` ? So then `  renderedComponent.click`, or just rendered ? But I understand that the type is `RenderedResult`, just thinking if I would understand what it means ?

timdeschryver: What about just `component`?
At the time I wrote this there was another variable named `component`, thus I could not use it...

tomastrajan: Yeah that would make sense too ;)

tomastrajan:
Also that codecov, that's strange, but haven't investigated yet ...

tomastrajan:
@timdeschryver merge whenever you're ready ;)

timdeschryver:
Yea, I was looking at codecov numbers but can't find a reason. Let's see how this evolves...

PR #412, opened 2018-11-01
Added docker-compose.yml to run the app using docker-compose

kosmas58:
## What:
Added docker-compose.yml to run the app using docker-compose.


tomastrajan:
@kosmas58 thank you for your contribution, can you please provide more background / links ? I am not very well versed with the docker technology. Thank you!

kosmas58:
@tomastrajan thanks for your reply.  To be honest, the benefits of docker-compose for Angular-Material-Starter are small. The main purpose of `docker-compose` is to orchestrate a series of containers (e.g. a web server with a database server) to generate and start them together.  I've added it because it's usually part of my toolset.

I've made the changes to the `Dockerfile` and `.dockerignore` because `husky` otherwise produces error messages under `docker run` and `docker-compose`.

tomastrajan:
@kosmas58 does this have an impact on people who would just want to run in with most basic docker setup? Do they have to have installed something more in their environment? 

In general we want stuff to work for the most people possible without getting too specific in one direction and as said before don't really know docker that well so that's why I am asking ;)

kosmas58:
@tomastrajan  [Compose](https://docs.docker.com/compose/overview/) is part of the normal docker installation. 

> _Compose is a tool for defining and running multi-container Docker applications. With Compose, you use a YAML file to configure your application’s services. Then, with a single command, you create and start all the services from your configuration._

The only thing you need to use it, is a `docker-compose.yml` file, which I've added.

kosmas58:
@tomastrajan  many thanks for merging it in.

PR #416, opened 2018-11-08
Updated to Angular Core (and related) 7.0.3

kosmas58:
## What:
Updated to Angular 7.0.3 and related


tomastrajan:
@kosmas58 thanks for the PR, we have to wait a but with merging until we fix broken e2e tests.

PR #420, opened 2018-11-10
 test(stocks): add valid and invalid stock symbol tests, closes #372

huyen1990:
## What:

Issue number: N/A

PR #422, opened 2018-11-10
test(form): add input, validation, autosave tests, closes #421

### LINKED ISSUE #421: Add Cypress tests for Form
Add Cypress tests for the [Form example](https://tomastrajan.github.io/angular-ngrx-material-starter#/examples/form).

I would like to work on this.

chauey:
## What:

Issue number: N/A


cypress/integration/form.spec.ts
```diff
@@ -0,0 +1,70 @@
+// const book_title = 'The Little Schemer';
```
timdeschryver: remove this

cypress/integration/form.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+describe('Form Actions', () => {
+  beforeEach(() => {
+    cy.visit('/#/examples/form');
+  });
+
+  it('should be able to input all values', () => {
+    cy.get('mat-slide-toggle[formcontrolname="autosave"]').click();
```
timdeschryver: I'm not really a big fan of using `formcontrolname` in tests, because it shouldn't really matter to the user how we're naming our form fields or which "technology" we use.

I do prefer using visual aspects of the form that an end-user can see, for example labels and placeholders. For an example you can take a look at the [existing CRUD tests](https://github.com/tomastrajan/angular-ngrx-material-starter/blob/master/cypress/integration/crud.spec.ts) or in https://github.com/tomastrajan/angular-ngrx-material-starter/pull/420.

cypress/integration/form.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    cy.get('mat-error').should('have.length', 5);
+    cy.get('form')
+      .should('contain', 'Username is required')
+      .should('contain', 'Password is required')
+      .should('contain', 'E-mail is required')
+      .should('contain', 'Birthday is required')
+      .should('contain', 'Description is required');
+
+    // email format
+    cy.get('form input[formcontrolname="email"]').type('a');
```
timdeschryver: I'm not sure if we should be this detailed on form errors, because the validators should already be covered by Angular and/or in our unit tests.

We can leave it here tho as an example, just wanted to point this out.
Thoughts @tomastrajan ?

cypress/integration/form.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+      'someusername'
+    );
+    cy.get('input[formcontrolname="password"]').type('somepassword');
+    cy.get('input[formcontrolname="email"]').type('some@email.com');
+    cy.get('input[formcontrolname="birthday"]').type('1/1/2000');
+    cy.get('textarea[formcontrolname="description"]').type('some description');
+    cy.get('mat-checkbox[formcontrolname="requestGift"] label').click();
+
+    // TODO: slider
+    cy.get('mat-slider[formcontrolname="rating"]').click();
```
timdeschryver: I think we should also click the `send` action here and we could test if the snackbar shows up.
Thoughts?

cypress/integration/form.spec.ts
timdeschryver: I would rather change the tests when the text changes, instead of using `id`'s.
Because here again, the user sees the texts and interacts with the application by using the text, not the id's. If there is no text "available", I would use a property `data-testid` instead of `id`.

This is based on Kent C. Dodds's articles, e.g.: https://blog.kentcdodds.com/making-your-ui-tests-resilient-to-change-d37a6ee37269

chauey: I made the selectors more generic. Other than the mat-error. ANy other feedback? thx

cypress/integration/form.spec.ts
chauey: Added code to test the snackbar notifications. Did not even notice them before!

cypress/integration/form.spec.ts
chauey: seems useful, and we can always delete it if it causes problems?

timdeschryver: You're right

cypress/integration/form.spec.ts
timdeschryver: I like it, thanks!

cypress/integration/form.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  it('should be able to input all values', () => {
+    cy.contains('Auto Save')
+      .click()
+      .get('input')
+      .should('be.checked');
+    cy.get('[placeholder="Username"]')
+      .type('someusername')
+      .should('have.value', 'someusername');
```
timdeschryver: any reason why we check these values?

tomastrajan:
@chauey I have done some commits to master in the meantime, could you please resolve merge conflicts so that I can marge your contribution? Thank you!

timdeschryver:
Thanks @chauey !

Issue #385, opened 2018-10-07
It will be good to have a Dockerfile for run the project inside docker container


shootermv:
## Describe the feature you'd like:
It will be good to have a Dockerfile for run the project inside docker container
Im newbie in docker and devops stuff, so if im right (and dockerfile it is good idea) - i will  be happy to try to do pr 

## I would be willing to submit a PR to fix this issue:

[ ] Yes (Assistance is provided if you need help submitting a pull request)  
[ ] No

tomastrajan:
@shootermv I also don't have much docker experience but if it is possible just to add a `Dockerfile` which enables users to run it in container while not influencing the base use case of running stuff with npm then I am in favor of such an addition :)

DinoSourcesRex:
@tomastrajan 
I've created a new `docker-compose` and `dockerfile` that seem production ready to me. It uses a multi-stage `dockerfile` to build the image for, hopefully, a quicker build on dev machines. Another benefit of the multi-stage files is that the final image should be much smaller then the build image as it only includes the vital stuff for serving. It should also let people easily host their site.

I have not managed to create a separate stage / image for running any testing. I had issues getting chrome or chrome headless working in there so I just didn't as I didn't need to do that locally.

---

The implementation raises a few issues, mainly that I have modified the `server.js` as well as creating 2 new docker related files (`Production.Dockerfile` and `Production.docker-compose.yml`).

I can raise a PR if this is something that is of interest, just let me know.

PS I also upgraded it to using node 11.1 with no issues that I saw at least.

tomastrajan:
@DinoSourcesRex please feel free to submit a PR :)

netlander:
Having problems deploying to Docker.

Where would the Docker scripts mentioned above be found?

PR #425, opened 2018-11-12
feat(docker): dockerfile for production

DinoSourcesRex:
## What:
* A `Production.Dockerfile` with 3 stages:
  * `npm_builder` so that docker can cache the `npm install` command so it is not re-run every time you make small code changes. 
  * `builder` which will run `build:prod`
  * `production` which is the final image and will take a fresh `node:11.1.0` image, copy the necessary files from the `builder` stage and then run `node server.js`
* A `Production.docker-compose.yml` which can be used to run and build the `Production.Dockerfile`. This can be run with `docker-compose -f Production.docker-compose.yml up`
* Updates to `server.js` which will take the port and context from environment variables, if those variables exist. Should not be a breaking change.
  * An update to remove the / and append it on CONTEXT instead so that I can control click the url from my VSCode terminal 😄 
---
Issue number: #385

### DIFF
--- a/Production.Dockerfile
+++ b/Production.Dockerfile
@@ -1,0 +1,35 @@
+FROM node:11.1.0 as npm_builder
+# Set the entrypoint as bin bash incase we want to inspect the container
+ENTRYPOINT ["/bin/bash"]
+# Manually copy the package.json
+COPY ./package.json /usr/src/app/package.json
+COPY ./package-lock.json /usr/src/app/package-lock.json
+COPY cypress /usr/src/app/cypress
+# Set the work directory to where we copied our source files
+WORKDIR /usr/src/app
+# Install all of our dependencies
+RUN npm install
+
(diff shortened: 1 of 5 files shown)


tomastrajan:
@DinoSourcesRex thank you very much for your contribution!

tomastrajan:
@DinoSourcesRex could you please also update README.md with relevant information for other users ?

DinoSourcesRex:
@tomastrajan Let me know how that looks. 

I can go into more depth on what can be changed by the docker-file if that's something you would like me to do (or add some comments into the dockerfile) - the main things I did not touch upon were changing the environment variables or the port it is being served on, which may be useful.

I've also added some npm scripts to run the production docker commands as they are a bit long-winded.

tomastrajan:
@DinoSourcesRex I think its pretty great thank you!

PR #428, opened 2018-11-16
refactor(styles): simplify the label's margin-bottom reset

ahasall:
Since the following PR https://github.com/twbs/bootstrap/pull/25561, there is a variable for setting the label's margin-bottom. We don't need to manually reset it.


tomastrajan:
@ahasall thanks for your contribution!

PR #429, opened 2018-11-17
fix: Remove non existing import NotificationStyles in error service

Doehl:
## What:

Remove a non existing import in the error handler service for NotificationStyles.

Can cause an error when typescript compiling the application.


tomastrajan:
Hi @Doehl ! 

Thank you for your contribution! You might considering also running `npm run contributors:add -- Doehl code` and `npm run contributors:generate` and commiting it to your PR to appear in the contributors list ;)

Cheers!

Doehl:
Hi tomatrajan,

Thanks, I have added my self as contributor

PR #430, opened 2018-11-17
test(app):  settings, examples and google-analytics tests, closes #419

### LINKED ISSUE #419: Write tests for untested reducers, actions and effects
I refactored some parts in https://github.com/tomastrajan/angular-ngrx-material-starter/pull/418 but forgot to write tests 😅 .
If someone wants to pick this one up feel free to do so, but leave a comment down below.
If no one picks this up during the next weeks, I will create them.

[x] Yes (Assistance is provided if you need help submitting a pull request)  
[ ] No

lau32:
Add tests for  settings actions and reducers, examples effects and google-analytics effects.

@timdeschryver, wasn't able to test the `changeLanguage` effect from the examples module.
Could you provide some guidance on that please?

### DIFF
--- a/src/app/core/google-analytics/google-analytics.effects.spec.ts
+++ b/src/app/core/google-analytics/google-analytics.effects.spec.ts
@@ -1,0 +1,49 @@
+import { getEffectsMetadata } from '@ngrx/effects';
+import { cold } from 'jasmine-marbles';
+import { GoogleAnalyticsEffects } from './google-analytics.effects';
+import { NavigationEnd } from '@angular/router';
+
+describe('GoogleAnalyticsEffects', () => {
+  let router: any;
+  const ga = (<any>window).ga;
+
+  beforeEach(() => {
+    router = {
+      routerState: {
+        snapshot: {}
+      },
+      events: {
+        pipe() {}
+      }
+    };
+
+    (<any>window).ga = jasmine.createSpy('ga');
+  });
+
+  afterAll(() => {
+    (<any>window).ga = ga;
+  });
+
+  it('should not dispatch action', function() {
+    const effect = new GoogleAnalyticsEffects(router);
+    const metadata = getEffectsMetadata(effect);
+
+    expect(metadata.pageView).toEqual({ dispatch: false });
+  });
+
+  it('should call google analytics', function() {
+    const routerEvent = new NavigationEnd(1, '', '');
+    router.events = cold('a', { a: routerEvent });
+    const effect = new GoogleAnalyticsEffects(router);
+
+    effect.pageView.subscribe(() => {
+      expect((<any>window).ga).toHaveBeenCalled();
+      expect((<any>window).ga).toHaveBeenCalledWith(
+        'set',
+        'page',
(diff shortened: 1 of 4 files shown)


tomastrajan:
@lau32 thanks for your contribution! Let's see what Tim has to say before merging though ;)

src/app/core/google-analytics/google-analytics.effects.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    expect(metadata.pageView).toEqual({ dispatch: false });
+  });
+
+  it('should call google analytics', function() {
+    const routerEvent = new NavigationEnd(1, '', '');
+    router.events = cold('a', { a: routerEvent });
+    const effect = new GoogleAnalyticsEffects(router);
+
+    effect.pageView.subscribe(() => {
+      expect((<any>window).ga).toHaveBeenCalled();
```
timdeschryver: Isn't this test a bit redundant, because we're also testing it with `toHaveBeenCalledWith` ?

src/app/examples/examples.effects.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      const effect = new ExamplesEffects(
+        actions,
+        store,
+        translateService,
+        router,
+        titleService
+      );
+
+      effect.setTitle.subscribe(() => {
+        expect(titleService.setTitle).toHaveBeenCalled();
```
timdeschryver: Same here.

timdeschryver:
Seems good to me!
What are the problems you encountered with the `changeLanguage` effect?

src/app/core/google-analytics/google-analytics.effects.spec.ts
lau32: Yes, that's true. I'll update that.

lau32:
Sorry, I meant to write `setTranslateServiceLanguage` effect.
It seams that if create call the `setTranslateServiceLanguage` with the store, after mocking the pipe on it like so: `store.pipe.and.returnValue(of(settings));`, the `translateService.use` method isn't being called.
Which is fair, because it will pass the entire pipe logic inside the examples effect.
But, how can I test that `setTranslateServiceLanguage` is being called when subscribing to `setTranslateServiceLanguage`?

timdeschryver:
Ah I see, I think mocking the `pipe` won't do much as you already are experiencing.
I would suppose the trigger would trigger if we change the store state, more specific the settings. Because `selectSettingsState` would execute and emit a new value.

What do you think, can you try this out?
If it isn't working you don't have to spend a lot of time on it, we'll merge this PR and I'll be taking a look for the `setTranslateServiceLanguage` test later.

tomastrajan:
@timdeschryver @lau32 what is the status here? Can I merge or is some changes still coming later?

lau32:
@tomastrajan you can merge this PR. I would look at another way to trigger the `setTranslateServiceLanguage` effect that would not require to create an instance of the store.

timdeschryver:
Thanks @lau32 !

PR #452, opened 2019-02-08
Update title.service.ts

DmitriyPotapov:
## What: Maybe in this case it is worth unsubscribing from observable?

Issue number: N/A

### DIFF
--- a/src/app/core/title/title.service.ts
+++ b/src/app/core/title/title.service.ts
@@ -8,6 +8,7 @@ import { environment as env } from '@env/environment';
 
 @Injectable()
 export class TitleService {
+  private unsubscribe$ = new Subject();
   constructor(
     private translateService: TranslateService,
     private title: Title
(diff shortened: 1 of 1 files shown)


timdeschryver:
I don't think this is needed because the the service exists as long as the application's life cycle.
Thoughts @tomastrajan ?

src/app/core/title/title.service.ts
```diff
@@ -27,8 +28,12 @@ export class TitleService {
       translate
         .get(title)
         .pipe(filter(translatedTitle => translatedTitle !== title))
-        .subscribe(translatedTitle =>
-          this.title.setTitle(`${translatedTitle} - ${env.appName}`)
+        .takeUntil(this.unsubscribe$)
```
timdeschryver: You'll have to use the pipeable operator here.

```
 .pipe(filter(translatedTitle => translatedTitle !== title), takeUntil(this.unsubscribe$)
)
```

src/app/core/title/title.service.ts
```diff
@@ -8,6 +8,7 @@ import { environment as env } from '@env/environment';
 
 @Injectable()
 export class TitleService {
+  private unsubscribe$ = new Subject();
```
timdeschryver: This isn't doing anything now.

In order to unsubscribe the observables, you will have to emit (and complete) the Subject's stream when the service is destroyed.

```
this.unsubscribe$.next();
this.unsubscribe$.complete();
```

tomastrajan:
@DmitriyPotapov I agree with the @timdeschryver  here, the service exists for the whole duration of the application life and so there is no reason to make things more complicated. Thanks for your effort and hope you enjoy the starter project!

PR #465, opened 2019-04-28
fix(prettier): run prettier on all subfolders and subfiles

abedzantout:
## What:

Prettier was running on the first three nested levels of the src,e2e and cypress folders only. This was due to not parsing the regex properly in package.json


tomastrajan:
@abedzantout thank you for the fix!

PR #469, opened 2019-06-18
refactor(path aliases): add core, shared and environments path aliases

alharto:
Updated the root tsconfig.json with 3 path aliases (shared, core and environments). Also updated several import paths to use said aliases as this makes refactoring the project structure easier.


tomastrajan:
We don't want this even though is nice but currently not supported by StackBlitz so yeah... Anyway thanks very much for your contribution and hope you enjoy the starter!

Issue #462, opened 2019-03-06
For getting version from package.json we are exporting entire package.json.


santoshyadav198613:
## Describe the feature you'd like:
Do not export entire package.json to get the version. It becomes the part of build,
## I would be willing to submit a PR to fix this issue:

[x] Yes (Assistance is provided if you need help submitting a pull request)  
[ ] No

tomastrajan:
@santoshyadav198613  any ideas on how to improve the current solution ?

santoshyadav198613:
Hi @tomastrajan ,
I will raise a PR sorry I forgot to do this.

I am traveling now, will raise in this weekend.

tomastrajan:
@santoshyadav198613 no worries ;) and than you!

santoshyadav198613:
Hi @tomastrajan ,
Check if the approach is fine, i am writing only the packages required to a version.json and than using it, so we can avoid exporting entire package.json

PR #474, opened 2019-07-02
fix(chore): take only version which is required

santoshyadav198613:
## What:

Issue number: #462

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,0 +1,46 @@
+/** Global imports */
+const gulp = require('gulp');
+const jsonEditor = require('gulp-json-editor');
+const packageJson = require('./package.json');
+
+/**  Gulp Tasks */
+
+/**
+ * project set up task. Default it local
+ *
+ * must pass env to setup prod like --env prod
+ *
+ * @example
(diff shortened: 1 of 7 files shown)


tomastrajan:
Hi @santoshyadav198613 ! Thanks for your contribution! 

The thing is I would love to keep number of dependencies / technologies as low as possible... This is for various reasons but mostly because it is a starter it should be as simple as possible and adding yet another build tool complicates things and can be confusing for the beginners...

Could this be achieved using npm scripts and maybe some inline / npx script or something inside of the `environment.ts` file itself?

santoshyadav198613:
> Hi @santoshyadav198613 ! Thanks for your contribution!
> 
> The thing is I would love to keep number of dependencies / technologies as low as possible... This is for various reasons but mostly because it is a starter it should be as simple as possible and adding yet another build tool complicates things and can be confusing for the beginners...
> 
> Could this be achieved using npm scripts and maybe some inline / npx script or something inside of the `environment.ts` file itself?

Yes, sure I will look for alternatives as well, as this was the first approach I had in my mind.

Issue #472, opened 2019-06-27
Use the NgRx 8 `create` factory functions


timdeschryver:
## Describe the feature you'd like:

The [release of NgRx 8](https://medium.com/ngrx/announcing-ngrx-version-8-ngrx-data-create-functions-runtime-checks-and-mock-selectors-a44fac112627) gave us a way to define actions and reducers in a less verbose way with `createAction` and `createReducer`. I would like to see this project to use these functions and the new `createEffect`.

## Other information:

I tried to list all of the features using NgRx here below. 
Feel free to claim your own feature 🙂

- [x] Auth
- [x] Settings
- [x] CRUD (Books)
- [x] Form
- [x] Stock market
- [x] Todos
- [x] google-analytics

We're here to guide you along the way.
(I will probably pick one feature in a couple of days, to get this started)

tomastrajan:
Also, here is a great blog post detailing how to do it in practice ! https://offering.solutions/blog/articles/2019/07/01/updating-your-angular-app-to-ngrx-8/

geromegrignon:
I'll work on CRUD part tonight.

tomastrajan:
@geromegrigon great, thanks ! 🔥

timdeschryver:
@evgenyfedorenko do you want to pick one up? Please don't feel obligated

evgenyfedorenko:
@timdeschryver a bit slammed at the moment with other stuff. I will certainly consider doing it later.

timdeschryver:
Sure, feel free to check in at any point 🙂

santoshyadav198613:
Hi @timdeschryver ,
I can take few, let me know which part i can work on.

timdeschryver:
I was just about to pick up `Auth` myself, feel free to claim your own @santoshyadav198613

santoshyadav198613:
Cool @timdeschryver , working on one very old issue, as soon as it is done, will pickup Todos.

santoshyadav198613:
I am taking Form next.

tomastrajan:
@timdeschryver  this is coming together in a very nice speed ! Amazing community work!

geromegrignon:
@abedzantout thanks for your contribution but for furthers contributions you should communicate about your contribution to avoid people working both on the same part.
@santoshyadav198613 : Form part has been merged fyi.

santoshyadav198613:
Cool @geromegrignon , in that case taking settings.

[...]

PR #475, opened 2019-07-02
refactor: use ngrx create functions for todo

santoshyadav198613:
## What: update store and effects to use create function for todo example.

Issue number: #472


tomastrajan:
@santoshyadav198613 thank you for your contribution!

PR #478, opened 2019-07-03
refactor(settings): refactored to ngrx 8

abedzantout:
## What:

Refactor settings ngrx functions to ngrx 8.

Issue number: https://github.com/tomastrajan/angular-ngrx-material-starter/issues/472

### DIFF
--- a/projects/angular-ngrx-material-starter/src/app/core/settings/settings.actions.spec.ts
+++ b/projects/angular-ngrx-material-starter/src/app/core/settings/settings.actions.spec.ts
@@ -1,90 +1,90 @@
 import {
-  ActionSettingsChangeAnimationsElements,
-  ActionSettingsChangeAnimationsPage,
-  ActionSettingsChangeAnimationsPageDisabled,
-  ActionSettingsChangeAutoNightMode,
-  ActionSettingsChangeHour,
-  ActionSettingsChangeLanguage,
-  ActionSettingsChangeStickyHeader,
-  ActionSettingsChangeTheme,
+  actionSettingsChangeAnimationsElements,
+  actionSettingsChangeAnimationsPage,
+  actionSettingsChangeAnimationsPageDisabled,
+  actionSettingsChangeAutoNightMode,
+  actionSettingsChangeHour,
+  actionSettingsChangeLanguage,
+  actionSettingsChangeStickyHeader,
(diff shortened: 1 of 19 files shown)


tomastrajan:
Hi @abedzantout ! Thank you for your contribution, the CI build seems to be red, can you have a look ? Thank you!

abedzantout:
@tomastrajan Done

projects/angular-ngrx-material-starter/src/app/core/settings/settings.actions.spec.ts
```diff
@@ -1,90 +1,90 @@
...
-  ActionSettingsChangeTheme,
+  actionSettingsChangeAnimationsElements,
+  actionSettingsChangeAnimationsPage,
+  actionSettingsChangeAnimationsPageDisabled,
+  actionSettingsChangeAutoNightMode,
+  actionSettingsChangeHour,
+  actionSettingsChangeLanguage,
+  actionSettingsChangeStickyHeader,
+  actionSettingsChangeTheme,
   SettingsActionTypes
```
tomastrajan: Hmm, this one ( `SettingsActionTypes` )was deleted in the file but is still imported here ?

timdeschryver: I think it still has to be deleted - https://github.com/tomastrajan/angular-ngrx-material-starter/blob/a3bc534412b4f1397cac898d922706c0a88f4b15/projects/angular-ngrx-material-starter/src/app/core/settings/settings.actions.ts#L5

abedzantout: Hmm, so you want to remove the SettingsActionTypes enum completely? We're using the same structure across almost all actions.

tomastrajan: As far as I understand, with new action creators there is  no need to have actions union type anymore ? cc @timdeschryver

timdeschryver: Yes, you can remove the enum completely.
Previously you had the enum to
 1) use it as the type to create the action
 2) use it in your reducers

Now, this becomes obsolete.

PR #483, opened 2019-07-25
fix(index): add missing </div>. refactor(scss): replace relative node_modules path with ~ operator

alharto:
## What: Add missing \</div> to index.html as it was causing async script loading at the end of the body tag to fail. Replace relative node_moldues paths with ~ operator is scss files.

Issue number: N/A


projects/angular-ngrx-material-starter/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": "src",
+    "paths": {
+      "@root/*": ["../../../*"],
+      "@app/*": ["app/*"],
+      "@core/*": ["app/core/*"],
+      "@shared/*": ["app/shared/*"],
+      "@assets/*": ["assets/*"],
+      "@environments/*": ["environments/*"]
```
tomastrajan: It was like this before but decided to move away from it cuz it breaks in stuff like `stackblitz` etc... So this will not be added

tomastrajan:
@alharto thank you for your contribution! As mentioned above the aliases we're removed for a reason so they will not be re-added for now. As for the div fix and scss, if it works for prod build we can do that!

projects/angular-ngrx-material-starter/tsconfig.json
alharto: Cool I'll revert that commit

alharto:
> @alharto thank you for your contribution! As mentioned above the aliases we're removed for a reason so they will not be re-added for now. As for the div fix and scss, if it works for prod build we can do that!

I've reverted that change

tomastrajan:
@alharto thank you! Just to make sure, did you try it locally with prod build and it was all running properly ?

alharto:
> @alharto thank you! Just to make sure, did you try it locally with prod build and it was all running properly ?

No worries thank you the starter project is great. Yes I ran the prod build and did a run through comparing against the live demo site everything looked good let me know if there are any issues if so I'll look at it ASAP.

Issue #495, opened 2019-10-29
Layout messed up in Edge with Animation query `position: fixed`


davideas:
## Minimal reproduction of the bug with instructions:
Just open the demo app in Edge: elements overlaps among them and layout is completely messed up. Cannot test on Safari, please verify yourself.

Line affected:
https://github.com/tomastrajan/angular-ngrx-material-starter/blob/425bdb831fabc70a03ef7f213b815a43994384dc/projects/angular-ngrx-material-starter/src/app/core/animations/route.animations.ts#L15

## Other information:
To fix in Edge, changing the style _position_ to `static` or `relative` will fix the layout and the scroll.
Because `position: fixed` helps a lot in Chrome and Firefox during the page transition: it skips pre-scrolling (or jumping), we should intercept if it is Edge at change the position value to something else:
``` typescript
query(':enter > *', style({ opacity: 0, position: isIEOrEdge ? 'relative' : 'fixed' }), {
```

dyingangel666:
Same happens in Safari. Whole layout is broken because there's still an inline style "position: fixed" from the animations ":enter" which is not removed from the DOM. I think it might be also possible that this is a bug with angular animation, because the prop from the animations was not removed from DOM. I haven't found a way to fix this.

dyingangel666:
Ok i've found a hacky fix, but it looks nicely. The problem is that some layouts, especially dynamic flex layouts are not working with absolute/fixed positioning of inner content for the router outlet. But when angular uses a router animation we have both ng-components injected to the dom the last page which is animating out, and the new one which is animating in. My layout is completely flexed and absolute positioning isn't possible. So my solution is to use a column flexed layout on the router-outlet wrapper container, and i've set the first following item of the router outlet to flex order: 2, and within the animation i switch the order, that when the leave animation is finished i show the the entering component at the top. For me this is working because my inner content has at least a height of available space from the header navbar to the bottom of the viewport.

Example animation (adaptable to the ngrx starter project):
`transition('* <=> *', [
    query(
      ':enter',
      [
        style({opacity: 0})
      ], {optional: true}
    ),
    sequence([
      query(
        ':leave',
        [
          style({opacity: 1}),
          animate(
            '1500ms ease-in',
            style({opacity: 0})
          ),
          style({order: 2})
        ], {optional: true}
      ),
      query(
        ':enter',
        [
          style({opacity: 0}),
          animate(
            '1500ms ease-in',
            style({opacity: 1})
[...]

Issue #493, opened 2019-10-11
UI issue. Un usable on mobile


clembou590:
Hello.
I wanted to start a website using this starter that looked perfect on my laptop.
However on my iphone 8. (ios 12.4)
with chrome 69.0.3497.91
(I do not see the issue on safari...)
Look at the picture: the website is not usable...

![iss](https://user-images.githubusercontent.com/10899818/66662294-d9f4a480-ec48-11e9-9ee9-7c7b0cd9ec40.png)

I tried also in incognito mode and got the same result....
Hopefully someone can fix it...?

tomastrajan:
Hi @clembou590 !

Thx,  for the feedback ! Interesting, can you please try to disable animations in the settings ? by passing it settings url directly so maybe that page is not too broken and then try to reload main page without settings ?

echol:
@tomastrajan Unfortunately I encountered this issue too. Just today I uploaded my own project that I was developing using your starter kit, this only happens on iPhone devices. It's such a shame cause this is really a good starter pack. I guess it's got something to do with CSS on iOS devices. Please update here if anyone finds any fix. I'll try and do some research myself.

clembou590:
@tomastrajan 
I disable the animations in the code in AnimationsService.ts
I wrote "NONE" and it worked...

So now how to solve the issue for real? (disabling the animation is not really a good way to handle this right...)

I also tried on the new iphone Xs of my brother and it has the same problem. I also got it on my ipad-pro...

tomastrajan:
@clembou590 as a first step you could add logic that disables animations only on affected devices / user agents but yeah I fully agree that a proper solution would be better... Hard to say what is the problem as it mostly works and uses standard angular animations and css, hmm...

davideas:
I see the effect is actually very similar in Edge too, I suspect the cause is the same described in #495 with `position: fixed`. The reason is unknown to me.

dyingangel666:
Reason comes from the routeAnimations. There's a prop "position: fixed" which is not removed for Safari and Microsoft Browsers. Why? I don't know, but it might be possible that this is an angular bug. I haven't found a way to fix it and keep the animations work as well.

dyingangel666:
PR is open ;) Tested Chrome, Firefox, Safari, Edge and IE11

PR #498, opened 2019-11-07
Fixed router transitions/animations

dyingangel666:
Fixed the animations for the router transitions which are broken in Safari and Edge because a inline style from the animation "position: fixed" was not properly removed.

Closes: #495
Closes: #493


tomastrajan:
Hi @dyingangel666 !

Thx for the contribution! Could you please follow [conventional commit message](https://github.com/tomastrajan/angular-ngrx-material-starter/blob/master/CONTRIBUTING.md#commit-message-guidelines) format ? And also it would be great if the PR was re based into single commit (for example using `rebase -i HEAD~<number-of-commits>` )

So in the end PR should be one commit which will look something like this `fix(core): route animations...`

Besides that can you please have a look on the [failing build](https://travis-ci.org/tomastrajan/angular-ngrx-material-starter/builds/608627422?utm_source=github_status&utm_medium=notification) (its just a prettier error, so formating, can be fixed using `npm run format:write` )?


Thank you!

PR #514, opened 2020-02-20
Form inputs

KadarH:
I think that the button save in the examples module ( form ) has to be disabled if the auto save is disabled or while the form is not valid .

```
if( autosave ) => disabled
else while( form not valid ) => disabled
=> enabled
```

The actual use :

`[disabled]="form.get('autosave').value`

My proposition : 

`[disabled]="form.get('autosave').value || !form.valid"`

### DIFF
--- a/projects/angular-ngrx-material-starter/src/app/features/examples/form/components/form.component.html
+++ b/projects/angular-ngrx-material-starter/src/app/features/examples/form/components/form.component.html
@@ -19,7 +19,7 @@
           </span>
           <div class="row">
             <mat-form-field class="col" [ngClass]="routeAnimationsElements">
-              <input matInput placeholder="{{ 'anms.examples.form.placeholder1' | translate }}"
+              <input matInput placeholder="{{ 'anms.examples.form.placeholder1' | translate }}*"
                 formControlName="username">
               <mat-error *ngIf="form.get('username').invalid">
                 {{ 'anms.examples.form.placeholder1' | translate }}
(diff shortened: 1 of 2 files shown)


tomastrajan:
Hi @KadarH  !

We want to be able to also safe invalid form, it's like work in progress. Maybe not ready yet but we dont want to lose it and be able to continue when we come back.

We only want to prevent SUBMIT of the form if its not valid.

Hope that makes sense!

KadarH:
Hi @tomastrajan 

So if we want to be able to also save invalid form, it is more logical that the inputs are not required.
In my opinion, required restrictions is used to tell the user that he can't save the form until the input is not empty.

Proposition : 

1 - No restrictions ( all input become not required ) => save form 
Or 
2 - The form cannot be saved until the inputs are filled ( unless the autoSave is activated )

What do you think ?
Regards

KadarH:
> Hi @tomastrajan
> 
> So if we want to be able to also save invalid form, it is more logical that the inputs are not required.
> In my opinion, required restrictions is used to tell the user that he can't save the form until the input is not empty.
> 
> **Proposition :**
> 
> * No restrictions ( all input become not required ) => save form
> 
> Or
> 
> * The form cannot be saved until the inputs are filled ( unless the autoSave is activated )
> 
> What do you think ?
> Regards

tomastrajan:
@KadarH maybe we should just rename `Save` button to `Save draft` because i think its a valid use case... I work on something big, it's not yet all valid, I need more data to finish the filling of the form, but I want to continue tomorrow, so I save the draft and continue tomorrow. Once I filled everything and form is valid I can finally submit it for real ?

jiddisch (approved):

PR #516, opened 2020-02-22
fix(change-hours): run change hours interval outside of angular zone

### LINKED ISSUE #511: Angular testability is unstable due to changeHours effect
## Minimal reproduction of the bug with instructions:
Load up the site and check Angular's built-in Testability to verify it stabilizes after requests are made. This can be checked via the console in your browser's Dev tools: 

`window.getTestabilities().forEach(testability => console.log(testability.isStable()))`

This should display `true` eventually after all angular requests are completed. However it is always `false`. I believe this issue is caused by the `changeHours` effect that sets an interval within angular scope (assumed). 

## Expected behavior:
Checking Angular's testability to stabilize should return true after all requests have been made and completed.

## Other information:
This can be resolved by removing the `changeHours` effect, but it'd be nice to keep it in. The effect is setting a pendingMacroTask in the testability, which never completes. Checking Angular's testability to be stable is helpful to verify that it's ok to proceed to the next step in Protractor and Selenium style E2E test cases.

[X] Yes (Assistance is provided if you need help submitting a pull request)  
[ ] No

dvargas46:
## What: Updated change hours effect interval to run outside of Angular zone. This should fix the testability not stabilizing issue without affecting the nature of the effect (to adjust the theme upon app load based on current time).

Closes #511 

Issue number: #511

### DIFF
--- a/projects/angular-ngrx-material-starter/e2e/protractor.conf.js
+++ b/projects/angular-ngrx-material-starter/e2e/protractor.conf.js
@@ -9,7 +9,7 @@ const { SpecReporter } = require('jasmine-spec-reporter');
  */
 exports.config = {
   allScriptsTimeout: 11000,
-  specs: ['./src/setup.ts', './src/**/*.e2e-spec.ts'],
+  specs: ['./src/**/*.e2e-spec.ts'],
   capabilities: {
     browserName: 'chrome',
     chromeOptions: {
--- a/projects/angular-ngrx-material-starter/e2e/src/setup.ts
+++ b/projects/angular-ngrx-material-starter/e2e/src/setup.ts
@@ -1,3 +1,0 @@
-import { browser } from 'protractor';
-
-browser.waitForAngularEnabled(false);
--- a/projects/angular-ngrx-material-starter/src/app/core/settings/settings.effects.spec.ts
+++ b/projects/angular-ngrx-material-starter/src/app/core/settings/settings.effects.spec.ts
@@ -5,6 +5,7 @@ import { Actions, getEffectsMetadata } from '@ngrx/effects';
 import { TestScheduler } from 'rxjs/testing';
 import { Store } from '@ngrx/store';
 import { of } from 'rxjs';
+import { NgZone } from '@angular/core';
 
 import {
   AnimationsService,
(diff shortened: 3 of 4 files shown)


projects/angular-ngrx-material-starter/src/app/core/utils/utils.ts
```diff
@@ -0,0 +1,44 @@
+import { Subscription, SchedulerLike } from 'rxjs';
+import { NgZone } from '@angular/core';
+
+class LeaveZoneScheduler implements SchedulerLike {
```
tomastrajan: This looks a bit excessive ? Why not just wrap the stream with `runOutsideOfZone` inline in the effects class?

dvargas46: @tomastrajan , hey thanks for the feedback. As far as know, wrapping the interval stream with `runOutsideAngular` won't be enough since the scheduler dispatches the action within the zone anyway. That's why this utils was needed to define our own SchedulerLike to pass into the interval observable, which handled scheduling the task in/out of the zone. It's discussed a bit [here](https://stackoverflow.com/questions/43121400/run-ngrx-effect-outside-of-angulars-zone-to-prevent-timeout-in-protractor), which is where this resolution was derived.

I'm not sure if there is another way around it using observables from what I saw and tried. Alternatively we could use pure js `setInterval` to dispatch the action instead. like this perhaps:
```
  hour = 0;
  changeHour = this.ngZone.runOutsideAngular(() => 
    setInterval(() => {
      const hour = new Date().getHours();
      if (hour !== this.hour) {
        this.hour = hour;
        this.ngZone.run(() => this.store.dispatch(actionSettingsChangeHour({ hour })));
      }
    }, 3_000)
  );
```

wouldn't need the utils class for this and would work just the same sticking it in the effects (even though it wouldn't 'exactly' be an effect).

I'm still new to this, so I could be missing something obvious. Anyway, lemme know what you think of the proposed solutions.

p.s. another good [discussion](https://stackoverflow.com/questions/52018881/why-would-i-use-rxjs-interval-or-timer-polling-instead-of-window-setinterval) about why using interval observable over setInterval has some advantages, all of which may not matter too much for this app besides composability and sticking to ngrx architectural designs. 🤔

tomastrajan: Hi @dvargas46 !

Thank you for very detailed explanation, in that case I would rather go with much simpler `setInterval` solution, less code is always better !

dvargas46: Sounds good, I have updated my PR to use the simpler approach then. Thanks!

tomastrajan:
@dvargas46 thanks for your contribution!

Issue #513, opened 2020-02-20
Use Angular Material TestHarness in tests


tomastrajan:
It would be great to migrate tests to use new [Angular Material Component Test Harness](https://material.angular.io/guide/using-component-harnesses) in the unit tests.

The migration should be mostly mechanical as it's basically replacing ad-hoc selectors with the selectors provided by the `@angular/cdk`.

Test can be run in watch mode using `npm run watch`...

tomastrajan:
Please write comment if you wanna migrate some component tests so that we can prevent duplicate effort

mucan:
Hi! I would like to give it a try.

tomastrajan:
@mcanoglu thank you for great contribution!

For others, there are still many components that can be migrated to use `TestHarness` and now you even have great example of how to do it by @mcanoglu !

yousafnawaz:
Hi @tomastrajan,  I would like to migrate some components to use . `https://material.angular.io/guide/using-component-harnesses` Angular Material Component Test Harness

AlirezaEbrahimkhani:
Hi @tomastrajan 
I'm very happy to be able to help with this migration and I'm starting and working on the crud.component.ts and after finishing the next component that I will work on, I will inform ...

PR #517, opened 2020-02-23
Migration Angular Material Test Harness

mcanoglu:
## What:

The tests for the FormComponent and the ToDo Component are migrated to the new material test harness. 

If these are OK, i would also migrate the other components. 

Issue number: #513


tomastrajan:
@mcanoglu thank you for your contribution!

PR #522, opened 2020-05-04
Added rel="noopener noreferrer" in anchor and alt in image tag

Zaky7:
### 1. Added rel="noopener noreferrer"
When a link opens a URL in a new tab with target="_blank", it is very simple for the opened page to change the location of the original page because the JavaScript variable window.opener is not null and thus "window.opener.location can be set by the opened page. This exposes the user to very simple phishing attacks.

To prevent pages from abusing window.opener, use rel=noopener on <a href=> to force its value to be null on the opened pages. With this in place, window.opener is null in Chrome 49+, Opera 36+, Firefox 52+, Desktop Safari 10.1+, and iOS Safari 10.3+. For older browsers, use "noreferrer". Cumulatively, rel="noopener noreferrer" is the safest way to mitigate this vulnerability.

### 3. Changed all-contributer-html.js to generate safe anchor tag and image element 

Issue number: N/A

### DIFF
--- a/.all-contributors-html.js
+++ b/.all-contributors-html.js
@@ -13,8 +13,8 @@ const content = fs.readFileSync(TARGET, 'utf8');
 const data = contributors
   .map(
     c => `
-  <a class="contributor" href="${c.profile}" target="_blank">
-    <img src="${c.avatar_url}">
+  <a class="contributor" href="${c.profile}" rel="noopener noreferrer" target="_blank">
+    <img alt src="${c.avatar_url}">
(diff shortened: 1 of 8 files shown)


tomastrajan:
@Zaky7 thank you very much for your contribution!

Zaky7:
> 
> 
> @Zaky7 thank you very much for your contribution!

@tomastrajan thank you 😄  for your humble reply. Nice project for beginners starting with ngrx and angular material :)

PR #529, opened 2020-06-24
fix(i18n): use the own languages for different languages to display

KingMario:
It's user-unfriendly for a user to find himself surrounded by unknown strange characters after he has accidentally chosen the Chinese language. The same applies to other languages.
It would be helpful to show at least his known language(s) in his known language(s).

## What:

Use their own languages for different languages to display.

For someone who does not understand Chinese, choosing his own language from such a list is like a lottery.

image

Issue number: N/A


KingMario:
Using its own language to display a language name is a common practice in the industry.

The language selector for [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web).

image

tomastrajan:
@KingMario thank you for yoru contribution!

PR #531, opened 2020-06-30
feat(form): form enhancement with trim and cleave

KingMario:
## What:

Form enhancement with trim and cleave

* add trim directive for username to trim value on input
* add a phone number field
* add cleave directive for phone number and birthday auto-format on typing

Issue number: N/A


KingMario:
Hi @tomastrajan ,

I'm not sure if this kind of new feature is welcome. The trim function was built-in for Angular.js 1.x, but it's missing in Angular. I think it's essential for data-binding.

And for some reason, the exact lines diff of the i18n json files are not very clearly displayed. There's only one line added for each json file, but the diff page shows each whole file in diff.

If you need me to add more demo in the examples or some description for the docs, please let me know.

Thanks.

PR #543, opened 2020-10-23
feat(tool): change to eslint In favor of tslint deprecation

SiddharthSh:
## What:
eslint has been introduced as part of tslint deprecation.

* code has been slightly modified to support the linting since eslint is more strict when converted.


tomastrajan:
Hey @SiddharthSh !

Thank you very much for your contribution!

Don't we need to remove tslint.json files and the dependency ?

SiddharthSh:
Hey @tomastrajan, good catch, have pushed latest by removing the tslint and incorporating the component and directive selector to the eslint rules. Also, RTL is being ignored in the codebase so I let it remain as is.
We cant remove dependency for tslint since the eslint plugin depends on it. :)

SiddharthSh:
Not sure why the Travis CI build is failing, running on my system seems fine :(

tomastrajan:
@SiddharthSh yeah really strange as initially your PR built just fine, also in `[rtl]` you just added comment, hard to imagine what's going wrong...

tomastrajan:
@SiddharthSh btw one more thing, maybe it would make sense to add eslint entry here https://tomastrajan.github.io/angular-ngrx-material-starter#/feature-list to highlight that this is already future proof ;)

SiddharthSh:
For future-readiness, I think yea better to showcase, so pushed another commit with the latest changes @tomastrajan

tomastrajan:
@SiddharthSh merged and released, thank you for your contribution!

PR #552, opened 2021-06-10
fix: base href

tino-tg:
## What:
fix: set base href with trailing slash for production build

- in order to use relative path urls, the base-href tag has to be set with a trailing slash, e.g. `<base href="some_path/" />`
- for more information see [Stackoverflow Answer](https://stackoverflow.com/a/26043021/13226740) 
- tldr see screenshot:
 
base_href

## Issue number: 
see discussion in previous [PR](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/551) 

## Wat is fixed:
assets path urls, such as the logo, the favicon or images on "about" page

## What was successfully testet as well:
i18n files, that are also delivered from "assets" folder

## More:
the hash-routing was not the problem, but it isn't needed anymore to make the app work on GH pages --> I could do another PR to change this, if you like

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "watch": "ng test --configuration=test --browsers ChromeHeadless --watch --reporters dots",
     "e2e": "ng e2e",
     "e2e:ci": "ng e2e",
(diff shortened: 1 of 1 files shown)


tomastrajan:
@tino-tg thank you very much for the fix.

The removal of hash, if it really works would be also appreciated!

PR #554, opened 2021-06-22
Modularizing Material and FortAwesome under Core

dimitrileite:
## What:

Modularizing Material

Issue number: N/A

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -507,6 +507,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "dimitrileite",
+      "name": "Dimitri de Jesus",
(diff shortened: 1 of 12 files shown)


dimitrileite:
Modularizing Material and FortAwesome

tomastrajan:
Why would we want to import more material from start ?

dimitrileite:
> Why would we want to import more material from start ?

When modularizing you can import Material in just one place. It is exported to the root. 
I have imported all material components into one module to make it easier to develop. You can delete the compontents not in use. 
I did the same thing for FortAwesome but for only Icons and features used.
Check it out.  It makes coding easier.

tomastrajan:
@dimitrileite yeah but that has a bad impact on tree-shaking / startup time right ?

dimitrileite:
You are right! Congrats for the great work!

PR #555, opened 2021-07-10
All project is UP to DATE! Updated.

dimitrileite:
## What:

Issue number: N/A


tino-tg:
Looks quite like the last [PR](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/554) **plus** adding jquery and bootstrap.js. @dimitrileite  Would you please describe what these changes would achieve?

dimitrileite:
> Looks quite like the last [PR](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/554) **plus** adding jquery and bootstrap.js. @dimitrileite Would you please describe what these changes would achieve?

All project is now updated to the most recent in almost all dependencies. 
Dried all FortAwesome to just one file.
Dried all Material to just one module. (It's quicker now)
Updated Husky. 
All working.

tino-tg:
> > Looks quite like the last [PR](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/554) **plus** adding jquery and bootstrap.js. @dimitrileite Would you please describe what these changes would achieve?
> 
> All project is now updated to the most recent in almost all dependencies.
> Dried all FortAwesome to just one file.
> Dried all Material to just one module. (It's quicker now)
> Updated Husky.
> All working.

But as you can see, the build is failing. Did you investigate this, @dimitrileite?

And there is still the open question why these are almost the same changes as in [your last PR](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/554) **plus adding jquery and bootstrap.js** 
Isn't it still the case that this would violate the basic principle of the module architecture (core/shared/feature-modules) *and* increase the initial startup time?

dimitrileite:
> > > Looks quite like the last [PR](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/554) **plus** adding jquery and bootstrap.js. @dimitrileite Would you please describe what these changes would achieve?
> > 
> > 
> > All project is now updated to the most recent in almost all dependencies.
> > Dried all FortAwesome to just one file.
> > Dried all Material to just one module. (It's quicker now)
> > Updated Husky.
> > All working.
> 
> But as you can see, the build is failing. Did you investigate this, @dimitrileite?
> 
> And there is still the open question why these are almost the same changes as in [your last PR](https://github.com/tomastrajan/angular-ngrx-material-starter/pull/554) **plus adding jquery and bootstrap.js**
> Isn't it still the case that this would violate the basic principle of the module architecture (core/shared/feature-modules) _and_ increase the initial startup time?


Hello to all! Good evening!

Firstly, this commit puts the project updated again.

Then, if you dive in Material concepts you would end up by having a Module to initialize all your Material Components, in this case, our project was having separated points that was instantiating those Components, one here, other there. Once Core is run in the root of the project we can have just one module to locate and import some or all Material Components, and it is inside Core still.

It does not increase the startup time. It is the same. I have just put some other material components for the sake of new comers that would use any other Material Component. It's almost all there that anyone'd need.

Our project also has bootstrap loaded. It is easier to cope with and if you want to deal with Material Components and Design, the best solution nowadays is bootstrap 5. And one of the ways to have them working properly on an Angular 12 project you would use this solution: This [link](https://www.itsolutionstuff.com/post/how-to-add-bootstrap-5-in-angular-12example.html). In that case quoting Hardik Savani: "we will also install bootstrap with jquery and popper js. so that way you can also import bootstrap css and bootstrap js function. So i think this will be best solution for you i think."

In the same way of thinking, working with FortAwesome is not that simple. For you to use one icon you would end up by importing one by one and placing them on the constructor of the Module you are thinking of using.

The way I implemented concords/agrees with the Main Angular Concept which is "Modularize/Componentalarize" the project.

I did not brake the project architecture and improved performance, readability, implementation time, understandability so on. As a full-stack developer, once I got in contact with this repo I had a steep curve to notice how all things would grip together. I just want to simplify that curve to others. 

This project has chances to succeed. And we should ktsat (keep things simple and tidy). 

Lastly, I take the chance now to honor all of the more than 40 contributors for the great work and purpose.

Thanks!

PR #559, opened 2021-08-07
fix(form): replace pass in storage stop duplicate requests from form …

### LINKED ISSUE #527: Perhaps better way to update ngrx on form.valuechanges()
I noticed in the current form.component.html the following code is used in the template to update the ngrx model:

  ```
<ng-container *ngIf="formValueChanges$ | async as updatedForm">
    {{update(updatedForm)}}
</ng-container>
```

I found that this triggers multiple and duplicate actions for even a single key entry on an input field. It also triggers the action on clicks within the form when data has not changed. I am wondering if this is by design or whether the duplicate actions are tolerated for the sake of the automatic unsubscribe by using the | async pipe.

If the duplication is tolerated for the sake of the automatic unsubscribe, I think I have a possible improvement. Instead of a function call within the template, the template would just contain the subscription:

```
<ng-container *ngIf="formValueChanges$ | async as updatedForm">
</ng-container>
```

but the component would provide for the action to be triggered in a tap operator of the formValueChanges$ observable:

```
ngOnInit() {
    this.formValueChanges$ = this.form.valueChanges.pipe(
      debounceTime(500),
      filter((form: Form) => form.autosave),
      tap(updatedForm => update(updatedForm))
    );
}
```

This seems to avoid duplicate actions. Thoughts? Any potential drawbacks to this solution?

### LINKED ISSUE #558: Storing password in state
I know this is meant only as an example for newer users, but in the example of the form you are storing the password in the application state. While that is a security risk in itself, the value is also persisted in local storage. Again, I know this is just a small demo but it would be great to promote security while we are at it! BTW I'm a huge fan of this architecture. My thoughts on this are to either pull password from state altogether, or if you're interested in teaching a little security while you're at it, we could possibly even hash the password before it enters state.

If I have time, I'd love to. I'll pull a clean branch down and have at it.

[X ] Yes (Assistance is provided if you need help submitting a pull request)  
[ ] No

millbj92:
…state, closes #527, closes #558

## What:

After reading issue #527 I started looking at the state a little more and realized they were indeed right,
so I fixed that. I also discovered that the password value is not being omitted or masked in state or local storage.
State isn't such a big deal, but localStorage would be a security concern. I chose not to go with hashing and salting as I don't want to bloat this package with another library. May want to put a small sidenote on the form that more work needs to be done for it to be used in an actual production scenario. Cheers!

Issue number: #527 
Issue number: #558


tomastrajan:
hi @millbj92  would love to merge if there were no conflicts ;)

tomastrajan:
Hi @millbj92 , any update on this ? :D THank you!

millbj92:
oof, I've been super busy! I'll get to this ASAP!

millbj92:
@tomastrajan I went ahead and got it done. Build seems to be breaking though. Looks like CI is having trouble installing chrome? Not sure it's anything I'd be able to help with. Requeuing  may solve.

tomastrajan:
@millbj92 thank you, gonna have a look 👍

tomastrajan:
@millbj92 looks like a problem in `package.json` / `package-lock.json` as it happens during `npm ci` and npm ERR! Unexpected token < in JSON at position 45

tomastrajan:
But gonna merge and fix on master, thank you!

millbj92:
Bah, I forgot. This is probably due to me using npm 7+. Thanks for fixing! I keep meaning to switch back to v6, but version 7 is so much faster :p

